PR #94, opened 2018-09-27
AIDA-374

ecurley-nextcentury:
Add overloaded method to AIF to add confidence to hypothesis.
Modify .shacl to allow confidence on hypothesis.

### DIFF
--- a/src/main/java/edu/isi/gaia/AIFUtils.kt
+++ b/src/main/java/edu/isi/gaia/AIFUtils.kt
@@ -578,7 +578,7 @@ object AIFUtils {
      * @return The hypothesis resource.
      */
     @JvmStatic
-    fun makeHypothesis(model: Model, hypothesisURI: String, hypothesisContent: Set<Resource>,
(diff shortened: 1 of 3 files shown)


ecurley-nextcentury:
@rgabbard, Thanks for the quick turn-around! I know you're busy, but I'd like to know whether this makes sense. The subgraph node can also have a confidence, but there's no way to add one through the library. Similarly, MutualExclusion has a confidence and a subraph. Should the confidence be removed from subgraph in the .shacl?

rgabbard:
@ecurley-nextcentury : I think you are correct that it makes more sense to mark confidence on whatever "contains" the sub-graph rather than the sub-graph itself, so removing confidence from the sub-graph in SHACL is good.

ecurley-nextcentury:
> @ecurley-nextcentury : I think you are correct that it makes more sense to mark confidence on whatever "contains" the sub-graph rather than the sub-graph itself, so removing confidence from the sub-graph in SHACL is good.

Excellent. I will do so and push the change to this PR. Thanks!

PR #100, opened 2018-10-05
Change resource example to pass ImagesToAIFTest.java

cedorman:
This passes the ImagesToAIFTest.java, since there is a new requirement to have a colon with something before it.


rgabbard:
Thanks for catching this (we added the colon requirement because it turns out colon-less IRIs are not legal and some user's downstream RDF systems were dying on them)

PR #103, opened 2018-11-19
Updated surefile plugin for fix test failure; added source plugin ver…

cedorman:
Solve build problem.  Unclear why this happened now, and has not appeared earlier.


nextcen-dgemoets:
Rebuilt everything from scratch, ensured that it downloaded the specified versions.  No issues.

PR #111, opened 2019-01-18
AIDA-514: Add imports to OntModel that we're already reading in

ecurley-nextcentury:
This should make testing run faster while the NIST servers are down


ecurley-nextcentury:
> I see that the Kotlin version didn't use constants for the preloaded ontologies, but I think it's nicer if all of those Strings are made into constants the way they are in the Java version. So you might take the opportunity to turn them and your new URIs into private constants in the companion object to match the Java version,
> 
> Not required for merge, just a suggestion. Runs sure look nicer without the stacktraces!

Good suggestion. However, as none of the strings were constants in the Kotlin version and I foresee this being the last edit that we'll make to the Kotlin version before we deprecate it, I didn't think it was worth the effort to make them (or any of the other strings) constants.

PR #112, opened 2019-01-23
Aida 500: Enhance Java AIF library post-conversion

nextcen-dgemoets:
Work performed:

- Used AIFOrderedTurtleWriter to ensure that the Java and Kotlin versions of AIF produce the same output from the validation tests.
- Used IntelliJ built-in coverage tools to reveal unused classes, methods, and individual lines of code (both Java and Kotlin).
- Consolidated loadModel and loadOntologyWithFriendlyError as per AIDA-447 code review.
- Converted getTypeAssertions and getConfidenceAssertions from Kotlin to Java.
- Added full Javadoc to AIFUtils.java and ValidateAIF.java.
- Updated some documentation as per Craig request (see comment in AIDA-500).

See the section of the [Java-Kotlin Confluence page](https://nextcentury.atlassian.net/wiki/spaces/AIDA/pages/472187228/AIF+Kotlin+to+Java+Conversion+AIDA-447+and+AIDA-500) entitled, “Some notes on the approach for and results of AIDA-500” for further information and details.

For code review, you may want to:

- Generate and review Javadoc for the impacted files...and perhaps evaluate if any other files should have Javadoc added.
- Run IntelliJ’s coverage on at least the Java ValidExamples and evaluate the results.
- Review the code for getTypeAssertions, getConfidenceAssertions, and loadModel.
- Evaluate whether com.ncc.verdi.input.util.annotation.HypothesisSubgraph#trackResource() can make use of AIFUtils.getTypeAssertions().
- Build VERDI against the AIDA-500 AIF jar.

Also, I think part of the resolution to AIDA-500 is deciding what classes/methods/etc. should be deprecated or removed from the Java AIF.  There are some things that are only used by the unused RPISeedlingOntologyMapper class , things we didn't migrate like imagesToAifTest and ColdStart2AidaInterchange, and a few things that weren't even used in the Kotlin classes.  But just because they’re not currently used in the tests doesn’t mean they’re without merit.  In some cases perhaps a test could be added.


rgabbard:
"Also, I think part of the resolution to AIDA-500 is deciding what classes/methods/etc. should be deprecated or removed from the Java AIF."

Both `imagesToAif` and `ColdStart2AidaInterchange` will to some degree still be maintained by ISI, at least through M18, so please keep anything in the API that either of those were using.

nextcen-dgemoets:
imagesToAif and ColdStart2AidaInterchange were not migrated to the Java AIF (com.ncc.aif package), but they remain available in the Kotlin AIF (edu.isi.gaia), both of which coexist in the same jar file.  The comment you cited is referring only to removing unused API from the Java AIF.  My understanding is that ColdStart will soon be marked @deprecated in this repo because it's not being supported/enhanced by Next Century (e.g., sentiment, NOT), but it will remain available to its users via the Kotlin AIF.  The README will be further updated to reflect this when ColdStart is deprecated.  Please let me know if this doesn't match your understanding.

rgabbard:
@nextcen-dgemoets : Right, but we will probably migrate `imagesToAif` and `ColdStart2AidaInterchange` to the Java API in order to keep up with any future changes NCC makes to those APIs as the program develops.  For that reason, it would at least be useful to know what things used only by those you all are considering removing.

nextcen-dgemoets:
IntelliJ reports that the following are not referenced in the project:
- The methods in AidaDomainOntologiesCommon (the class is used only by Kotlin's PassThroughOntologyMapper, which is used by ColdStart).  Perhaps the private members (e.g., canHaveName) can simply be made public static members.
- The RPISeedlingOntologyMapper class-- not used by Java or Kotlin.
- ValidateAIF's validateTA3, ensureEveryNamedNodeHasARdfType, and the LACKS_TYPES_QUERY String member.  These are used in neither the Java nor the Kotlin version of AIF.
- AIFUtils' addStandardNamespaces, getTypeAssertions, getConfidenceAssertions, a few image-related methods, and the UuidIriGenerator class. With the exception of getTypeAssertions and getConfidenceAssertions, these are all used in Kotlin-only classes like imagesToAifTest and ColdStart2AidaInterchange.

So we can leave PassThroughOntologyMapper and everything in AIFUtils intact in the Java AIF.

The main things to consider removing, then, are those currently not used by Java or Kotlin (e.g., RPISeedlingOntologyMapper, ValidateAIF's items above, getTypeAssertions, and getConfidenceAssertions).

rgabbard:
> So we can leave PassThroughOntologyMapper and everything in AIFUtils intact in the Java AIF.

Great, thanks!

`RPISeedlingOntologyMapper` can definitely go, as can the things in `ValidateAIF`.

src/main/java/com/ncc/aif/ValidateAIF.java
```diff
@@ -122,8 +122,12 @@ public static void main(String[] args) throws IOException {
         boolean allValid = true;
         for (File fileToValidate : filesToValidate) {
             logger.info("Validating " + fileToValidate);
-            final Model dataToBeValidated = loadModel(Files.asCharSource(fileToValidate, Charsets.UTF_8)
-                    .openBufferedStream());
+            final Model dataToBeValidated = ModelFactory.createOntologyModel();
+            try {
+                loadModel(dataToBeValidated, Files.asCharSource(fileToValidate, Charsets.UTF_8));
+            } catch (Exception e) {
```
ecurley-nextcentury: If you're unable to load the file to be validated, do you want the validation to continue anyway? If so, we shouldn't catch the RuntimeException.

src/main/java/com/ncc/aif/ValidateAIF.java
```diff
@@ -139,86 +143,41 @@ public static void main(String[] args) throws IOException {
     }
 
     /**
-     * Returns whether or not the KB and hypotheses are valid
-     */
-    public boolean validateTA3(Model dataToBeValidated) {
```
ecurley-nextcentury: If we're removing this method, we should probably also remove the ta2ShaclModel and the underlying aida_hypothesis.shacl file

src/main/java/com/ncc/aif/AIFUtils.java
```diff
@@ -130,18 +172,26 @@ public static Resource makeRelationInEventForm(Model model, String relationUri,
...
+     * @param system   The system object for the system which created this event
+     * @return The created event resource
      */
     public static Resource makeEvent(Model model, String eventUri, Resource system) {
         return makeAIFResource(model, eventUri, AidaAnnotationOntology.EVENT_CLASS, system);
     }
 
     /**
-     * Marks an entity as filling an argument role for an event or relation.
+     * Mark an entity as filling an argument role for an event or relation.
```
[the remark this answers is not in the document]
nextcen-dgemoets: Added.

src/main/java/com/ncc/aif/AIFUtils.java
```diff
@@ -152,8 +202,15 @@ public static Resource markAsArgument(Model model, Resource eventOrRelation, Res
...
```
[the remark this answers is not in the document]
nextcen-dgemoets: Added.

src/main/java/com/ncc/aif/AIFUtils.java
```diff
@@ -195,6 +257,16 @@ public static Resource markType(Model model, String typeAssertionUri, Resource e
...
+     * Make a justification.
+     *
+     * @param model      The underlying RDF model for the operation
+     * @param docId      The String document Id of the source of the justification
+     * @param classType  The class of the justification (e.g., text, image, video)
+     * @param system     The system object for the system which made this justification
+     * @param confidence The confidence with which to mark the justification
+     * @return The created justification resource
+     */
     public static Resource makeAIFJustification(Model model, String docId, Resource classType,
```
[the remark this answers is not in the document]
nextcen-dgemoets: Definitely not-- good catch!  As per quick convo, I converted JavaDoc to a single-line comment now that it's private.
I noticed that SparqlQueries should also be private and made a similar change.

src/main/java/com/ncc/aif/ValidateAIF.java
nextcen-dgemoets: I think at the time I was thinking that it should log the failure but continue the batch validations, but the original Kotlin terminates, so I changed this one to do the same.

src/main/java/com/ncc/aif/ValidateAIF.java
nextcen-dgemoets: Removed.

PR #118, opened 2019-03-01
AIDA-572

ecurley-nextcentury:
Remove entity/event restriction requiring justifications to be from the same source

### DIFF
--- a/src/main/resources/com/ncc/aif/restricted_aif.shacl
+++ b/src/main/resources/com/ncc/aif/restricted_aif.shacl
@@ -9,39 +9,6 @@
 #-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#
 
 #########################
-# All Entity and Event objects are restricted to justifications
-# from a single document; that is, there can be multiple justification
-# sections, but need to have the same aida:source.
-#------------------------
-aida:EntityShape
-    sh:sparql [
-        sh:message "Entity justifications can only have 1 source" ;
-        sh:select """
-            PREFIX aida:  <https://tac.nist.gov/tracks/SM-KBP/2018/ontologies/InterchangeOntology#>
-            SELECT $this
-            WHERE {
-                $this aida:justifiedBy [ aida:source|aida:containedJustification/aida:source ?source ] .
-            }
-            GROUP BY $this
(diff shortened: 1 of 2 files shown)


ecurley-nextcentury:
Yeah, I set FORCE_DUMP to false so that we don't dump the model by default. Let me know if you think that's a mistake.

I'm not sure sure what the value of JIRA's "relates to" flag is, but I'm eager to find out ;).

I removed the restriction that Craig mentioned in the issue from the internal page. As Craig isn't actively updating the Collaboration NIST restrictions page, I'm following his lead. Let me know if he mentioned something else to you. 

As you say, we're going to wait on the email. Craig mentioned that we should wait until NIST provides the restrictions in the eval plan. 

Let me know if there's anything else. If I don't hear anything by Monday morning, I'll go ahead and merge this

nextcen-dgemoets:
Yes, I agree with the change in FORCE_DUMP.  Saves me from having to do it locally all the time.

I'll be happy to extol the virtues of "relates to" ad nauseum to you someday (just kidding)  In general, I think it can help link two issues that may otherwise not be obviously linked (and encourage you to check out an issue if you're looking at something that relates to it), but in this case when they are all part of the same User Story, I think the benefit is minimal.

I'll let Craig weigh in on when to update AIDAC's version of the validator page.

The PR is ready to merge from my perspective.

craigwarsaw:
> I'll let Craig weigh in on when to update AIDAC's version of the validator page.
If I understand correctly, I don't think it is worth updating https://nextcentury.atlassian.net/wiki/spaces/AIDAC/pages/505053485/AIF+Validator
That page doesn't speak about specific restrictions, so I wouldn't want to introduce that precedence just for this.
I think the original JIRA ticket was referring to removing it from:
https://nextcentury.atlassian.net/wiki/spaces/AIDA/pages/396820524/Pilot+AIF+Restrictions+and+Hypotheses
and
https://nextcentury.atlassian.net/wiki/spaces/AIDAC/pages/353796390/Pilot+Evaluation+Information+Including+AIF+Restrictions+and+Hypotheses
I think it is best to leave those pages as-is and just make sure the M18 versions of these pages are accurate (once NIST releases the M18 restrictions).

PR #121, opened 2019-03-07
AIDA-609: Add LDC Ontologies to repo

ecurley-nextcentury:


ecurley-nextcentury:
Good catch! I have added the missing relations to the ontologies.

I'm not really sure what "author_check" is. It isn't an entity type and only appears for:
Unable to add constraint (author_check) for argument Evaluate.Deliberateness.Accidental_Holder.
Unable to add constraint (author_check) for argument Evaluate.Deliberateness.Deliberate_Holder.
Unable to add constraint (author_check) for argument Evaluate.Legitimacy.Illegitimate_Holder.
Unable to add constraint (author_check) for argument Evaluate.Legitimacy.Legitimate_Holder.
Unable to add constraint (author_check) for argument Evaluate.Sentiment.Negative_Holder.
Unable to add constraint (author_check) for argument Evaluate.Sentiment.Positive_Holder.

Perhaps it's something particular to 'Evaluate'? Currently it's being ignore, but we can special-case it. I would recommend we talk to LDC about it first.

PR #124, opened 2019-03-13
AIDA-550

ecurley-nextcentury:
Remove object property restriction for Relation/Event argument assertions

### DIFF
--- a/src/main/resources/com/ncc/aif/aida_ontology.shacl
+++ b/src/main/resources/com/ncc/aif/aida_ontology.shacl
@@ -187,7 +187,7 @@ aida:EventArgumentShape
     sh:message "The subject of an event argument must be an event"
   ] ;
 
-  # the object is the entity participating in the event
+  # the object is the argument of the event role
   sh:property [
     sh:path rdf:object ;
     sh:xone ( [sh:class aida:Entity ]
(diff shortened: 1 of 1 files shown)


ecurley-nextcentury:
I wanted to keep them there so when we address them later, we know what we did before. In fact, given the evaluation plan, these should probably be restricted to entities/events/relations.

ecurley-nextcentury:
@nextcen-dgemoets, I've updated this to allow entities/events/relations. Please review.

src/main/resources/com/ncc/aif/aida_ontology.shacl
```diff
@@ -190,7 +190,10 @@ aida:EventArgumentShape
   # the object is the entity participating in the event
```
craigwarsaw: the comment is out of date since it only references an entity

src/main/resources/com/ncc/aif/aida_ontology.shacl
```diff
@@ -250,8 +253,11 @@ aida:RelationArgumentShape
...
```
craigwarsaw: the comment is out of date since it only references an entity. Plus it looks like a copy/paste error because it references an event but this is in a RelationArgumentShape

PR #126, opened 2019-03-19
Aida 655

ecurley-nextcentury:
Add sourceDocument to justifications

### DIFF
--- a/src/main/java/com/ncc/aif/AIFUtils.java
+++ b/src/main/java/com/ncc/aif/AIFUtils.java
@@ -362,13 +362,12 @@ public class AIFUtils {
     /**
      * Add a sourceDocument to a pre-existing justification
      *
-     * @param model              The underlying RDF model for the operation
      * @param justification      A pre-existing justification resource
      * @param sourceDocument     A string containing the source document id
      * @return The modified justification resource
      */
-    public static Resource addSourceDocumentToJustification(Model model, Resource justification, String sourceDocument) {
-        justification.addProperty(AidaAnnotationOntology.SOURCE_DOCUMENT, model.createTypedLiteral(sourceDocument));
+    public static Resource addSourceDocumentToJustification(Resource justification, String sourceDocument) {
+        justification.addProperty(AidaAnnotationOntology.SOURCE_DOCUMENT, sourceDocument);
         return justification;
     }
 
--- a/src/main/java/com/ncc/aif/AIFUtils.java
+++ b/src/main/java/com/ncc/aif/AIFUtils.java
@@ -293,7 +293,7 @@ public class AIFUtils {
      * Create a justification from a particular snippet of text.
      *
      * @param model              The underlying RDF model for the operation
-     * @param docId              The String document Id of the source of the justification
+     * @param docId              A string containing the document element (child) ID of the source of the justification
      * @param startOffset        An integer offset within the document for the start of the justification
      * @param endOffsetInclusive An integer offset within the document for the end of the justification
      * @param system             The system object for the system which made this justification
@@ -325,7 +325,7 @@ public class AIFUtils {
      *
      * @param model              The underlying RDF model for the operation
      * @param toMarkOn           The Resource to be marked by the specified text document
-     * @param docId              The String document Id of the source of the justification
+     * @param docId              A string containing the document element (child) ID of the source of the justification
      * @param startOffset        An integer offset within the document for start of the justification
      * @param endOffsetInclusive An integer offset within the document for the end of the justification
      * @param system             The system object for the system which marked this justification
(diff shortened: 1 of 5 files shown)


src/main/resources/com/ncc/aif/aida_ontology.shacl
```diff
@@ -610,88 +610,110 @@ aida:SystemPropertyShape
...
-   sh:property [
-       sh:path aida:endOffsetInclusive ;
-       sh:or ( [sh:datatype xsd:int] [sh:datatype xsd:integer] )  ;
-       sh:minCount 1;
-       sh:maxCount 1 ];
-   # an optional string label for user consumption (TODO: is this allowed? Probably not?)
-   sh:property [
-      sh:path skos:prefLabel ;
-      sh:datatype xsd:string ];
+    # each justification must have exactly one source document ID
```
nextcen-dgemoets: Suggest "...exactly one source document element ID".  The good news is that with `aida:SharedJustificationShape`, this only needs to be done once.  :)

src/main/resources/com/ncc/aif/aida_ontology.shacl
```diff
@@ -610,88 +610,110 @@ aida:SystemPropertyShape
...
+    sh:property [
+        sh:path aida:source ;
+        sh:datatype xsd:string ;
+        sh:minCount 1;
+        sh:maxCount 1 ] ;
 
-   # justifications must provide confidences
-   sh:property aida:RequiredConfidencePropertyShape ;
+    # allow sourceDocument property as well
```
nextcen-dgemoets: Suggest something like, "allow souceDocument property for source document (parent) Id."

src/main/resources/com/ncc/aif/ontologies/InterchangeOntology
```diff
@@ -245,6 +245,13 @@
   rdfs:range xsd:string ;
   rdfs:subPropertyOf owl:topDataProperty .
 
+:sourceDocument rdf:type owl:DatatypeProperty ;
+  rdfs:label "source document" ;
+  rdfs:comment "Specify xsd:string identifier for source document of justification" ;
```
nextcen-dgemoets: Suggest "[...] source element of justification" for `aida:source` and possibly "[...] source parent document of justification" for `aida:sourceDocument`.

src/test/java/com/ncc/aif/ExamplesAndValidationTest.java
```diff
@@ -564,6 +564,7 @@ void createCompoundJustification() {
 
             markJustification(personTypeAssertion, textJustification);
             markJustification(putin, textJustification);
+            addSourceDocumentToJustification(textJustification, "NYT_PARENT_ENG_20181231_03");
```
nextcen-dgemoets: I wonder if it would be more "interesting" to use this new API in an example with two documents, such as relationWhereBothEndpointsAreAmbiguousCrossDocEntitiesVersion() or relationWhereBothEndpointsAreAmbiguousNISTRestrictedVersion().  I guess since these are just the unrestricted AIF examples, we'd want to find a case where there's some utility to adding a sourceDocument, like where there are some document elements that belong to more than one document.  Anyway, we don't absolutely need a change here, but I thought we might be able to find a more interesting example to which to add the sourceDocument.  LMK what you think.

src/main/java/com/ncc/aif/AIFUtils.java
```diff
@@ -358,6 +359,18 @@ public static Resource markTextJustification(Model model, Collection<Resource> t
         return justification;
     }
 
+    /**
+     * Add a sourceDocument to a pre-existing justification
+     *
+     * @param justification      A pre-existing justification resource
+     * @param sourceDocument     A string containing the source document id
```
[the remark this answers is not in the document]
ecurley-nextcentury: fixed

src/main/resources/com/ncc/aif/aida_ontology.shacl
ecurley-nextcentury: fixed

src/main/resources/com/ncc/aif/aida_ontology.shacl
ecurley-nextcentury: fixed

src/main/resources/com/ncc/aif/ontologies/InterchangeOntology
ecurley-nextcentury: fixed

ecurley-nextcentury:
Yeah, I agree. It's also nice that it's just comments, rather than the name of the property. That way, we can change the convention if we need to later (hopefully not necessary).

src/test/java/com/ncc/aif/ExamplesAndValidationTest.java
ecurley-nextcentury: I agree that it would be more interesting, but I'm not sure it would be any more helpful. NIST requires that all justifications have this extra property and this shows the researchers how to use it. In fact, all the tests with justifications that use the nist validator will have to change when that restriction is implemented

PR #134, opened 2019-03-21
Aida 671

ecurley-nextcentury:
This is dependent on #133, so feel free to look at this later. Do let me know if you have any questions/concerns.

### DIFF
--- a/src/main/java/com/ncc/aif/ValidateAIF.java
+++ b/src/main/java/com/ncc/aif/ValidateAIF.java
@@ -154,13 +154,13 @@ public final class ValidateAIF {
     // Show usage information.
     private static void showUsage() {
         System.out.println("Usage:\n" +
-                "\tvalidateAIF { --ldc | --program | --ont FILE ...} [--nist | --hypo] [-h | --help] {-f FILE ... | -d DIRNAME}\n" +
+                "\tvalidateAIF { --ldc | --program | --ont FILE ...} [--nist] [--nist-ta3] [-h | --help] {-f FILE ... | -d DIRNAME}\n" +
                 "Options:\n" +
                 "--ldc\t\tValidate against the LDC ontology\n" +
                 "--program\t\tValidate against the program ontology\n" +
                 "--ont FILE ...\tValidate against the OWL-formatted ontolog(ies) at the specified filename(s)\n" +
                 "--nist\t\tValidate against the NIST restrictions\n" +
-                "--hypo\t\tValidate against the NIST hypothesis restrictions (assumes --nist)\n" +
+                "--nist-ta3\t\tValidate against the NIST hypothesis restrictions (implies --nist)\n" +
                 "-h, --help\tShow this help and usage text\n" +
                 "-f FILE ...\tvalidate the specified file(s) with a .ttl suffix\n" +
                 "-d DIRNAME\tValidate all .ttl files in the specified directory\n" +
@@ -209,7 +209,7 @@ public final class ValidateAIF {
                 case "--nist":
                     flags.add(ArgumentFlags.NIST);
                     break;
-                case "--hypo":
+                case "--nist-ta3":
                     flags.add(ArgumentFlags.NIST);
                     flags.add(ArgumentFlags.HYPO);
                     break;
@@ -372,7 +372,7 @@ public final class ValidateAIF {
         if (restriction == Restriction.NIST) {
             logger.info("-> Validating against NIST SHACL.");
         } else if (restriction == Restriction.NIST_HYPOTHESIS) {
-            logger.info("-> Validating against NIST Hypothesis SHACL");
+            logger.info("-> Validating against NIST Hypothesis SHACL.");
         }
         logger.info("*** Beginning validation of " + filesToValidate.size() + " file(s). ***");
 
--- a/src/main/java/com/ncc/aif/ValidateAIF.java
+++ b/src/main/java/com/ncc/aif/ValidateAIF.java
@@ -28,8 +28,12 @@ import java.util.*;
  */
 public final class ValidateAIF {
 
+    // Enum to track restrictions
+    public enum Restriction {NONE, NIST, NIST_HYPOTHESIS}
+
     private static final String AIDA_SHACL_RESNAME = "com/ncc/aif/aida_ontology.shacl";
     private static final String NIST_SHACL_RESNAME = "com/ncc/aif/restricted_aif.shacl";
+    private static final String NIST_HYPOTHESIS_SHACL_RESNAME = "com/ncc/aif/restricted_hypothesis_aif.shacl";
     private static final String INTERCHANGE_RESNAME = "com/ncc/aif/ontologies/InterchangeOntology";
     private static final String AIDA_DOMAIN_COMMON_RESNAME = "com/ncc/aif/ontologies/AidaDomainOntologiesCommon";
     private static final String LDC_RESNAME = "com/ncc/aif/ontologies/SeedlingOntology";
@@ -41,28 +45,33 @@ public final class ValidateAIF {
 
     private static Model shaclModel;
     private static Model nistModel;
+    private static Model nistHypoModel;
     private static boolean initialized = false;
 
     private static void initializeSHACLModels() {
         if (!initialized) {
             shaclModel = ModelFactory.createOntologyModel();
-            CharSource aifSource = Resources.asCharSource(Resources.getResource(AIDA_SHACL_RESNAME), Charsets.UTF_8);
-            loadModel(shaclModel, aifSource);
+            loadModel(shaclModel, Resources.asCharSource(Resources.getResource(AIDA_SHACL_RESNAME), Charsets.UTF_8));
 
             nistModel = ModelFactory.createOntologyModel();
-            loadModel(nistModel, aifSource);
+            nistModel.add(shaclModel);
             loadModel(nistModel, Resources.asCharSource(Resources.getResource(NIST_SHACL_RESNAME), Charsets.UTF_8));
+
+            nistHypoModel = ModelFactory.createOntologyModel();
+            nistHypoModel.add(nistModel);
+            loadModel(nistHypoModel,
+                    Resources.asCharSource(Resources.getResource(NIST_HYPOTHESIS_SHACL_RESNAME), Charsets.UTF_8));
 
             initialized = true;
         }
     }
 
     private Model domainModel;
-    private boolean useRestrictedAIF;
-    private ValidateAIF(Model domainModel, boolean nistFlag) {
+    private Restriction restriction;
+    private ValidateAIF(Model domainModel, Restriction restriction) {
         initializeSHACLModels();
         this.domainModel = domainModel;
-        this.useRestrictedAIF = nistFlag;
+        this.restriction = restriction;
     }
 
     // Ensure what file name an RDF syntax error occurs in is printed, which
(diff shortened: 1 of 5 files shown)


src/test/java/com/ncc/aif/ExamplesAndValidationTest.java
```diff
@@ -1394,86 +1470,73 @@ void invalid() {
...
-
-                markJustification(entity, compound1);
-                markJustification(relation, compound1);
-                markJustification(event, compound1);
-
-                assertAndDump(model, "NIST.invalid: CompoundJustification must be used only for " +
-                                "justifications of argument assertions",
-                        nistSeedlingValidator, false);
-
+                testValid("NIST.invalid: Cluster has IRI");
```
nextcen-dgemoets: Should be testInvalid.

src/test/java/com/ncc/aif/ExamplesAndValidationTest.java
```diff
@@ -1207,32 +1208,119 @@ void missingRdfTypeOnNamedNode() {
...
-        Model model;
-        Resource system;
+    class NISTExamples extends BaseExample {
         Resource entity;
         Resource event;
+        Resource relation;
         Resource entityCluster;
         Resource eventCluster;
+        Resource relationCluster;
+        NISTExamples() { validator = nistSeedlingValidator; }
```
nextcen-dgemoets: NISTExamples could @override `testValid` and `testInvalid` to prepend `"NIST.valid"` or `"NIST.invalid"` so we could remove it from the calls to `testValid` and `testInvalid` and promote uniformity and avoid copy-paste errors.

Just a suggestion for your consideration-- I don't feel strongly.

src/test/java/com/ncc/aif/ExamplesAndValidationTest.java
```diff
@@ -1207,32 +1208,119 @@ void missingRdfTypeOnNamedNode() {
...
+
+        // Each edge justification must be represented uniformly in AIF by
+        // aida:CompoundJustification, even if only one span is provided
+        // Edges are assumed to be relation and event arguments
+        //
+        // CompoundJustification must be used only for justifications of argument assertions,
+        // and not for justifications for entities, events, or relation KEs
+        @Nested
+        class RestrictJustification {
+            @Test
```
nextcen-dgemoets: It looks like you moved `RestrictJustification` to the top of the class definition; it used to be at the bottom after `ConfidenceValueRange`.  Was this deliberate, and if so, why?  It just causes a bunch of additional diffs, but if there's a reason, then no worries.

Also, I'm not really reviewing the content of `RestrictJustification` and `ClusterHasIRI` although in theory I should, since we're re-adding it to the tree.

src/test/java/com/ncc/aif/ExamplesAndValidationTest.java
```diff
@@ -1207,32 +1208,119 @@ void missingRdfTypeOnNamedNode() {
         }
     }
 
+    class BaseExample {
```
sharknc: I would add a java doc or comment here describing the purpose of this class and what is used for.

src/test/java/com/ncc/aif/ExamplesAndValidationTest.java
```diff
@@ -1394,86 +1470,73 @@ void invalid() {
...
+        @BeforeEach
+        void setup() {
+            super.setup();
+            entity = makeEntity(model, getEntityUri(), system);
+            addType(entity, SeedlingOntology.Person);
+            entityCluster = makeClusterWithPrototype(model, getClusterUri(), entity, system);
+        }
 
-                markJustification(eventEdge, compound);
+        // Exactly 1 hypthesis should exist in model
```
nextcen-dgemoets: typo: hypothesis

src/test/java/com/ncc/aif/ExamplesAndValidationTest.java
```diff
@@ -1394,86 +1470,73 @@ void invalid() {
...
-                        ImmutableSet.of(justification1),
-                        system,
-                        1d);
-
-                markJustification(relationEdge, compound);
+    /**
+     * Set of tests to show that NIST Hypothesis restrictions pass and fail appropriately
+     */
+    @Nested
+    class NISTHypothesisExamples extends BaseExample {
```
nextcen-dgemoets: There's some voice in my head that insists that NISTHypothesisExamples should subclass NISTExamples which should subclass BaseExample.  But there's really no reason it should right now and NISTHypothesisExamples (at least at this point) doesn't need some of the extra stuff that NISTExamples has.  But because of the NIST restriction, "Each hypothesis graph must have at least one event or relation with at least one edge," (Section 4.3 - TA3 req #5) that suggests we're going to need them soon.  Maybe we should just do it now?

src/test/java/com/ncc/aif/ExamplesAndValidationTest.java
```diff
@@ -1207,32 +1208,119 @@ void missingRdfTypeOnNamedNode() {
...
+    }
+
     /**
      * Set of tests to show that NIST restrictions pass and fail appropriately
      */
     @Nested
-    class NISTExamples {
-        Model model;
-        Resource system;
+    class NISTExamples extends BaseExample {
```
sharknc: As a group @ecurley-nextcentury , @nextcen-dgemoets, @sharknc we should decide if we like this design and if we want to move forward with extending all nested test classes with BaseExample. Currently NISTExample is the only class that extends BaseExample. My vote is to keep it, but just update the rest of the test classes to make everything uniform.

src/test/java/com/ncc/aif/ExamplesAndValidationTest.java
```diff
@@ -1394,86 +1470,73 @@ void invalid() {
...
 
-                // test event argument
-                final Resource eventEdge = markAsArgument(model, event, SeedlingOntology.Conflict_Attack_Target, entity, system, 1.0);
+        @BeforeEach
+        void setup() {
+            super.setup();
+            entity = makeEntity(model, getEntityUri(), system);
+            addType(entity, SeedlingOntology.Person);
+            entityCluster = makeClusterWithPrototype(model, getClusterUri(), entity, system);
+        }
```
nextcen-dgemoets: I guess `NISTHypothesisExamples` would also have to @override `testValid` and `testInvalid` if you took my suggestion from before.

src/test/java/com/ncc/aif/ExamplesAndValidationTest.java
```diff
@@ -1207,32 +1208,119 @@ void missingRdfTypeOnNamedNode() {
...
+            model = createModel();
+            system = makeSystemWithURI(model, getTestSystemUri());
+        }
+
+        void addType(Resource resource, Resource type) {
+            markType(model, getAssertionUri(), resource, type, system, 1d);
+        }
+
+        void testInvalid(String name) {
+            assertAndDump(model, name, validator, false);
```
sharknc: Please update the java doc / comments for assertAndDump() on line 1546. The current wording sounds like the model is dumped for every test. After talking with @ecurley-nextcentury the model is only getting dumped when a test fails or if the FORCE_DUMP flag is set to _true_.

src/test/java/com/ncc/aif/ExamplesAndValidationTest.java
sharknc: This should be calling the `testInvalid()` method.

src/test/java/com/ncc/aif/ExamplesAndValidationTest.java
```diff
@@ -1394,86 +1470,73 @@ void invalid() {
...
 
-                markJustification(eventEdge, compound);
+        // Exactly 1 hypthesis should exist in model
+        @Nested
+        class SingleHypothesis {
+            @Test
+            void invalidTooMany() {
+                makeHypothesis(model, getUri("hypothesis-1"), Collections.singleton(entity), system);
+                makeHypothesis(model, getUri("hypothesis-2"), Collections.singleton(entity), system);
+                testInvalid("NISTHypothesis.invalid (too many): there should only be 1 hypothesis");
```
nextcen-dgemoets: Suggest "exactly one hypothesis" here and below.

src/main/java/com/ncc/aif/ValidateAIF.java
```diff
@@ -154,13 +154,13 @@ public static ValidateAIF create(ImmutableSet<CharSource> domainOntologySources,
     // Show usage information.
```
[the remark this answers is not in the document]
nextcen-dgemoets: Y'know, I did a quick search for something to help with parsing args, but I didn't come across picocli unfortunately.  The two solutions I found were over-engineered and involved adding another large-ish third-party library dependency just for one CLI...didn't seem worth it.  Picocli seems like it might be more in-line with our needs.  I'll write a JIRA ticket to examine it for use in the validator.  Feel free to edit/comment it after it's created with your thoughts.

src/test/java/com/ncc/aif/ExamplesAndValidationTest.java
nextcen-dgemoets: I agree, and moreover I think that ultimately we'll want NISTHypothesisExamples to subclass NISTExamples because soon they'll need events and relations.

src/test/java/com/ncc/aif/ExamplesAndValidationTest.java
ecurley-nextcentury: I think this is a good idea, but I think it should probably be incorporated into the next refactoring of this file. Ideally, we'd like to split this file up and I think I'd like the implementation of this idea to have the individual tests extend from a AbstractBaseTestSet or something common that would run testValid/Invalid. The concrete subclasses could then be responsible for producing a valid and invalid model. We can work out the details later, but let me know what you think

src/test/java/com/ncc/aif/ExamplesAndValidationTest.java
ecurley-nextcentury: I wanted to keep similar tests together. This is now at the top to be near EdgeJustificationLimit. If you feel that is too confusing, I'm OK to move it back

nextcen-dgemoets: No not at all-- I just wanted to make sure there was some kind of reason and it wasn't some kind of side effect of merge-error-fixing.  Works for me.

src/test/java/com/ncc/aif/ExamplesAndValidationTest.java
ecurley-nextcentury: I'm not sure how that would work with @Nested tests. I'm inclined to see how much will be shared and parcel it out as appropriate. Concerns with that approach?

nextcen-dgemoets: Well, because of 4.3 `TA3 req #5`  we're going to need the `entity`, `event`, `relation`, `entityCluster`, `eventCluster`, and `relationCluster` members from `NISTExamples` and the `setup()` that goes with it.  We'd need to reassign the `validator` variable.

But if there are complications with @Nested tests we can at least defer the topic until we implement new hypothesis restrictions.

ecurley-nextcentury: Well, as a hypothesis only requires a single event or relation edge, I don't think they'd need all those things. I guess they'd avoid either the event+cluster or relation+cluster. Luckily, it sounds like JUnit has us covered: https://www.petrikainulainen.net/programming/testing/junit-5-tutorial-writing-nested-tests/. In light of this, I am removing BaseExamples and putting the functionality into the top-level class

src/test/java/com/ncc/aif/ExamplesAndValidationTest.java
nextcen-dgemoets: I don't have a problem with it being part of the refactoring you describe (which I like), but sometimes refactoring tasks get set aside (not necessarily on AIDA but in general).  Would you object to it being written up as a JIRA task and set as a blocker for [AIDA-611](https://nextcentury.atlassian.net/browse/AIDA-611) (which is currently in our next Sprint)?  That means the new issue would need to go in the next Sprint also.

src/test/java/com/ncc/aif/ExamplesAndValidationTest.java
nextcen-dgemoets: Funny, I was just looking at that page.  Sounds like a good approach.

src/test/java/com/ncc/aif/ExamplesAndValidationTest.java
ecurley-nextcentury: Depends on when we get to a stable ExampleAndValidationTest.java. I would like to avoid refactoring the single file into multiples while there are several people working on it at once. That being said, it _should_ be a ticket and blocking 611 is a really good place for it.

src/main/java/com/ncc/aif/ValidateAIF.java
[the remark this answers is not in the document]
nextcen-dgemoets: Wrote AIDA-696 to capture this.

src/test/java/com/ncc/aif/ExamplesAndValidationTest.java
nextcen-dgemoets: Okay, we can defer this until we have a better handle of the scope of hypothesis tests (which are probably still at least a bit in flux).  And we can weigh that against any downside to subclassing.

src/test/java/com/ncc/aif/ExamplesAndValidationTest.java
nextcen-dgemoets: The most recent commit added Javadoc to assertAndDump() explaining this.

src/test/java/com/ncc/aif/ExamplesAndValidationTest.java
nextcen-dgemoets: BaseExample was removed in [`d72c1`](https://github.com/NextCenturyCorporation/AIDA-Interchange-Format/pull/134/commits/d72c1df86a351b1aed6e80ad7a548e69ea454427) in favor of moving its functionality to the ExamplesAndValidationTest class.

src/test/java/com/ncc/aif/ExamplesAndValidationTest.java
nextcen-dgemoets: This class has been removed as part of [`d72c1`](https://github.com/NextCenturyCorporation/AIDA-Interchange-Format/pull/134/commits/d72c1df86a351b1aed6e80ad7a548e69ea454427), and its functionality moved to the `ExamplesAndValidationTest` class.

src/test/java/com/ncc/aif/ExamplesAndValidationTest.java
nextcen-dgemoets: Added [AIDA-698](https://nextcentury.atlassian.net/browse/AIDA-698) to capture this.

PR #141, opened 2019-03-25
Aida 691

bdpham:
Implemented add_source_document_to_justification in aifutils.py and updated tests.


python/aida_interchange/aifutils.py
```diff
@@ -318,6 +318,12 @@ def mark_compound_justification(g, things_to_justify, justifications, system, co
     return compound_justification
 
 
+def add_source_document_to_justification(g, justification, source_document) :
+    g.add((justification, AIDA_ANNOTATION.sourceDocument,
```
ecurley-nextcentury: It's a public API, so we should add a comment

PR #142, opened 2019-03-26
AIDA-692

jakevb:
Added python implementation of java code from AIDA-654.


python/aida_interchange/aifutils.py
```diff
@@ -332,9 +332,10 @@ def make_cluster_with_prototype(g, cluster_uri, prototype, system):
     """
     cluster = _make_aif_resource(g, cluster_uri, AIDA_ANNOTATION.SameAsCluster, system)
     g.add((cluster, AIDA_ANNOTATION.prototype, prototype))
+    if handle == None:
```
bdpham: I know I'm not put as a reviewer, but shouldn't this be !=?

jakevb: Yes it should.  Typo.  Thanks.

python/aida_interchange/aifutils.py
```diff
@@ -332,7 +332,7 @@ def make_cluster_with_prototype(g, cluster_uri, prototype, system, handle=None):
     """
     cluster = _make_aif_resource(g, cluster_uri, AIDA_ANNOTATION.SameAsCluster, system)
     g.add((cluster, AIDA_ANNOTATION.prototype, prototype))
-    if handle == None:
+    if handle != None:
```
bdpham: The preferred Python syntax is "is not None"

sharknc: @jakevb I believe you can explicitly state "is not None" like @bdpham suggested: `if handle is not None:`

jakevb: Yep, I changed that in the most recent push.

python/tests/Examples.py
```diff
@@ -568,6 +570,33 @@ def test_simple_hypothesis_with_importance_with_cluster(self):
 
         self.dump_graph(g, "Simple hypothesis with importance with cluster")
 
+    def create_a_simple_cluster_with_handle(self):
```
bdpham: Without "test" as the first word in the test function name, unit test will skip this function/test.

python/tests/Examples.py
```diff
@@ -568,6 +570,33 @@ def test_simple_hypothesis_with_importance_with_cluster(self):
 
         self.dump_graph(g, "Simple hypothesis with importance with cluster")
 
+    def test_create_a_simple_cluster_with_handle(self):
```
bdpham: This test is throwing an Exception.`Exception: term 'handle' not in namespace 'https://tac.nist.gov/tracks/SM-KBP/2018/ontologies/InterchangeOntology#'`
The look up table in aida_rdf_ontologies.py SEEDLING_TYPES_NIST needs to have the term "handle" added.

python/aida_interchange/aifutils.py
```diff
@@ -332,9 +332,10 @@ def make_cluster_with_prototype(g, cluster_uri, prototype, system):
     """
     cluster = _make_aif_resource(g, cluster_uri, AIDA_ANNOTATION.SameAsCluster, system)
     g.add((cluster, AIDA_ANNOTATION.prototype, prototype))
+    if handle is not None:
+        g.add((cluster, AIDA_ANNOTATION.handle, handle))
```
bdpham: In rdflib, the object in a triple (handle) needs to be a literal.

jakevb:
Sorry, merge conflicts.  Could you re-review?

PR #145, opened 2019-03-27
AIDA 704:  Clusters must be able to contain Relations

nextcen-dgemoets:
Work performed:
Reflect the fact that relations can be added to clusters in InterchangeOntology, ontology README, and AIFUtils Javadoc.
Deprecate AIFUtils.makeRelationInEventForm() and update examples not to use it.

I hedged my bets a bit in the README by saying "should be able to" because we don't actually use it / test it.  Should I make it stronger by saying it's "unsupported" or something?  Do we support it?

### DIFF
--- a/src/main/java/com/ncc/aif/AIFUtils.java
+++ b/src/main/java/com/ncc/aif/AIFUtils.java
@@ -160,7 +160,10 @@ public class AIFUtils {
      * @param system           The system object for the system which created the specified relation
      * @param confidence       If non-null, the confidence with which to mark the specified relation
      * @return The created relation resource
+     * @deprecated AIF no longer represents relations as simply a connection between two entities,
+     * so use {@link #makeRelation(Model, String, Resource)} instead
      */
+    @Deprecated
     public static Resource makeRelationInEventForm(Model model, String relationUri, Resource relationType, Resource subjectRole,
                                                    Resource subjectResource, Resource objectRole, Resource objectResource,
                                                    String typeAssertionUri, Resource system, Double confidence) {
@@ -261,7 +264,7 @@ public class AIFUtils {
 
     // Helper function to create a justification (text, image, audio, etc.) in the system.
     private static Resource makeAIFJustification(Model model, String docId, Resource classType,
-                                                Resource system, Double confidence) {
+                                                 Resource system, Double confidence) {
         final Resource justification = makeAIFResource(model, null, classType, system);
         justification.addProperty(AidaAnnotationOntology.SOURCE, model.createTypedLiteral(docId));
         markConfidence(model, justification, confidence, system);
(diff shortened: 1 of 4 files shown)


src/main/resources/com/ncc/aif/ontologies/README.md
```diff
@@ -21,7 +21,8 @@ a hand-written source file and is not generated.
 
 ### InterchangeOntology
 
-This ontology contains the Interchange vocabulary and how the defined classes and properties may interact.
+This ontology contains the Interchange vocabulary and specifies how the defined classes and properties may interact.
+It is defined in such a way that it should be able to be used with OWL reasoners.
```
ecurley-nextcentury:
```suggestion
It is defined using strict OWL and should be able to be used with OWL reasoners.
```

src/main/java/com/ncc/aif/AIFUtils.java
```diff
@@ -160,7 +160,10 @@ public static Resource makeRelation(Model model, String relationUri, Resource sy
      * @param system           The system object for the system which created the specified relation
      * @param confidence       If non-null, the confidence with which to mark the specified relation
      * @return The created relation resource
+     * @deprecated AIF no longer represents relations as simply a connection between two entities,
```
ecurley-nextcentury:
```suggestion
     * @deprecated This method doesn't allow the user access to the blank nodes created for type and argument assertions. Use {@link #makeRelation(Model, String, Resource)} instead
```

src/test/java/com/ncc/aif/ExamplesAndValidationTest.java
```diff
@@ -1397,7 +1390,7 @@ void valid() {
                 // setup() already makes type assertions on the entity, event,
                 // and relation objects, justified by a text justification.
 
-                // Note that if you use makeRelationInEventForm, you will need to use the Jena API to obtain
+                // Note that if you use the deprecated makeRelationInEventForm, you will need to use the Jena API to obtain
```
ecurley-nextcentury: Should we even mention this? Will this encourage users to look at makeRelationInEventForm instead of ignoring it (as they should)?

nextcen-dgemoets: Yeah, I was wondering that myself.  Removed.

src/main/resources/com/ncc/aif/ontologies/README.md
nextcen-dgemoets: Changed.

src/main/java/com/ncc/aif/AIFUtils.java
nextcen-dgemoets: Much better, thanks.  I also noticed it uses the same confidence for all assertions, which might not be what you want.

PR #147, opened 2019-03-28
Aida 673

sharknc:
- Created hypothesis restriction to enforce edge KE's in the hypothesis graph must have exactly one edge importance value
- Created hypothesis restriction to enforce event and relation clusters in the hypothesis must have exactly one importance value
- Created hypothesis restriction to enforce that the hypothesis must have exactly one importance value
- Created valid and invalid test cases to validate all new aforementioned hypothesis restrictions 
- Updated all other hypothesis test cases that were impacted by the required importance value

### DIFF
--- a/src/main/resources/com/ncc/aif/restricted_hypothesis_aif.shacl
+++ b/src/main/resources/com/ncc/aif/restricted_hypothesis_aif.shacl
@@ -25,6 +25,46 @@ aida:SystemShape
         """ ;
     ] .
 
+#########################
+# 4.3 TA3 #4 Each hypothesis graph must have exactly one hypothesis importance value
+#------------------------
+aida:HypothesisGraphImportanceRequiredShape
+    a sh:PropertyShape ;
+    sh:path ( aida:HypothesisShape aida:ImportancePropertyShape ) ;
+    sh:minCount 1 ;
+    sh:maxCount 1 .
+
+#########################
+# 4.3 TA3 #5 Each event or relation (cluster) in the hypothesis must have exactly one event/relation importance value
+#------------------------
+aida:HypothesisEventRelationClusterImportanceRequiredShape
+    a sh:NodeShape ;
+
+    sh:target [
+        a sh:SPARQLTarget ;
+        sh:select """
+            PREFIX aida:  <https://tac.nist.gov/tracks/SM-KBP/2018/ontologies/InterchangeOntology#>
+            SELECT ?prototype
+            WHERE {
+                ?prototype a aida:SameAsCluster .
+                ?prototype aida:prototype/a ?type .
+                FILTER( ?type = aida:Event || ?type = aida:Relation )
+            }rest
+        """ ;
+    ];
+
+    sh:property [
+        a sh:PropertyShape ;
+        sh:path aida:importance ;
+        sh:minCount 1 ;
+        sh:maxCount 1 ;
+    ] .
+
+#########################
+# 4.3 TA3 #6 Each edge KE in the hypothesis graph must have exactly one edge importance value
+#------------------------
+
+
 ########################
 # 4.3 TA3 #7 Each entity (cluster) in the hypothesis graph must
 # have exactly one text description (called a "handle")
--- a/src/main/resources/com/ncc/aif/restricted_hypothesis_aif.shacl
+++ b/src/main/resources/com/ncc/aif/restricted_hypothesis_aif.shacl
@@ -25,37 +25,21 @@ aida:SystemShape
         """ ;
     ] .
 
-#########################
+#########################.;
 # 4.3 TA3 #4 Each hypothesis graph must have exactly one hypothesis importance value
 #------------------------
 aida:HypothesisImportanceRequiredShape
     a sh:PropertyShape ;
-    sh:path aida:importance;
+    sh:path aida:importance ;
+    sh:or (
+        [sh:propertyShape aida:HypothesisShape]
+        [sh:propertyShape aida:EventArgumentShape]
+        [sh:propertyShape aida:RelationArgumentShape]
+    ) ;
     sh:minCount 1 ;
     sh:maxCount 1 .
 
 aida:HypothesisShape
-    sh:property aida:HypothesisImportanceRequiredShape .
-
-#########################
-# 4.3 TA3 #5 Each event or relation (cluster) in the hypothesis must have exactly one event/relation importance value
-#------------------------
-aida:HypothesisEventRelationClusterImportanceRequiredShape
-    a sh:NodeShape ;
-
-    sh:target [
-        a sh:SPARQLTarget ;
-        sh:select """
-            PREFIX aida:  <https://tac.nist.gov/tracks/SM-KBP/2018/ontologies/InterchangeOntology#>
-            SELECT ?prototype
-            WHERE {
-                ?prototype a aida:SameAsCluster .
-                ?prototype aida:prototype/a ?type .
-                FILTER( ?type = aida:Event || ?type = aida:Relation )
-            }
-        """ ;
-    ];
-
     sh:property aida:HypothesisImportanceRequiredShape .
 
 #########################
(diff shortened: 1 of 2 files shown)


src/main/resources/com/ncc/aif/restricted_hypothesis_aif.shacl
```diff
@@ -25,6 +25,59 @@ aida:SystemShape
         """ ;
     ] .
 
+#########################.;
+# 4.3 TA3 #4 Each hypothesis graph must have exactly one hypothesis importance value
+#------------------------
+aida:HypothesisImportanceRequiredShape
+    a sh:PropertyShape ;
+    sh:path aida:importance ;
+    sh:or (
```
ecurley-nextcentury: Why include this?

src/main/resources/com/ncc/aif/restricted_hypothesis_aif.shacl
```diff
@@ -25,6 +25,59 @@ aida:SystemShape
...
+
+    sh:target [
+        a sh:SPARQLTarget ;
+        sh:select """
+            PREFIX aida:  <https://tac.nist.gov/tracks/SM-KBP/2018/ontologies/InterchangeOntology#>
+            SELECT ?prototype
+            WHERE {
+                ?prototype a aida:SameAsCluster .
+                ?prototype aida:prototype/a ?type .
+                FILTER( ?type = aida:Event || ?type = aida:Relation )
```
ecurley-nextcentury: I was concerned that there would be more overhead with at FILTER than with a UNION. Let me know what you think about this alternative:
`{ ?prototype aida:prototype/a aida:Event } UNION { ?prototype aida:prototype/a aida:Relation }`

src/main/resources/com/ncc/aif/restricted_hypothesis_aif.shacl
```diff
@@ -25,6 +25,59 @@ aida:SystemShape
...
+            }
+        """ ;
+    ];
+
+    # sh:maxCount 1 is inherited from aida_ontology.shacl
+    sh:property [
```
ecurley-nextcentury: Why not just use the property shape defined above? Also, this looks like aida:importance lives on the cluster. Is that what's expected. The issue says something about an "event/relation importance importance value".

src/main/resources/com/ncc/aif/restricted_hypothesis_aif.shacl
```diff
@@ -25,6 +25,59 @@ aida:SystemShape
         """ ;
     ] .
 
+#########################.;
+# 4.3 TA3 #4 Each hypothesis graph must have exactly one hypothesis importance value
+#------------------------
+aida:HypothesisImportanceRequiredShape
```
ecurley-nextcentury: This is used for more than hypothesis importance
```suggestion
aida:ImportanceRequiredShape
```

src/test/java/com/ncc/aif/ExamplesAndValidationTest.java
```diff
@@ -1508,12 +1516,162 @@ void invalidMultipleHandles() {
...
+                testValid("NISTHypothesis.valid: Each hypothesis graph must have exactly one" +
+                        " hypothesis importance value");
+            }
+        }
+
+        // Each event or relation (cluster) in the hypothesis must have exactly one importance value
+        @Nested
+        class HypothesisEntityRelationClusterImportanceValue {
+
+            private final String documentEventUri = getUri("ldc:event-1");
```
ecurley-nextcentury: Get tacks the string on the end of http://www...ldc#. You probably don't want to use ldc:, etc.

src/test/java/com/ncc/aif/ExamplesAndValidationTest.java
```diff
@@ -1508,12 +1516,162 @@ void invalidMultipleHandles() {
...
+            private final String relationUri = getUri("ldc:relation-1");
+            Resource event;
+            Resource relation;
+
+            @BeforeEach
+            void setup() {
```
ecurley-nextcentury: I like the way this class builds up from invalid to valid. Maybe that's a paradigm that would work well in solving Craig's problem

src/test/java/com/ncc/aif/ExamplesAndValidationTest.java
```diff
@@ -1508,12 +1516,162 @@ void invalidMultipleHandles() {
...
+                testValid("NISTHypothesis.valid: Each event or relation (cluster) in the hypothesis must " +
+                        "have exactly one importance value");
+            }
+        }
+
+        // Each edge KE in the hypothesis graph must have exactly one edge importance value
+        @Nested
+        class HypothesisEdgeImportanceValue {
+
+            private final String documentEventUri = getUri("ldc:event-1");
```
ecurley-nextcentury: Same comment as above: should just use "event-1"

src/test/java/com/ncc/aif/ExamplesAndValidationTest.java
```diff
@@ -1508,12 +1516,162 @@ void invalidMultipleHandles() {
...
+
+                markImportance(makeClusterWithPrototype(model, getClusterUri(), event, system), 88);
+                markImportance(makeClusterWithPrototype(model, getClusterUri(), relation, system), 88);
+            }
+
+            @Test
+            void invalidEventEdge() {
+
+                //invalid event argument, needs importance value
+                markAsArgument(model, event, SeedlingOntology.Personnel_Elect_Elect,
```
ecurley-nextcentury: You might want to add a URI for the argument assertions so that your invalid messages will be easier to read. This applies to relations as well

src/test/java/com/ncc/aif/ExamplesAndValidationTest.java
```diff
@@ -1508,12 +1516,162 @@ void invalidMultipleHandles() {
...
+                markImportance(makeHypothesis(model, getUri("hypothesis-1"),
+                        Collections.singleton(entity), system), 100);
+                testValid("NISTHypothesis.valid: Each hypothesis graph must have exactly one" +
+                        " hypothesis importance value");
+            }
+        }
+
+        // Each event or relation (cluster) in the hypothesis must have exactly one importance value
+        @Nested
+        class HypothesisEntityRelationClusterImportanceValue {
```
ecurley-nextcentury:
```suggestion
        class HypothesisEventRelationClusterImportanceValue {
```

src/main/resources/com/ncc/aif/restricted_hypothesis_aif.shacl
sharknc: Great point. I originally did not have this in here but I had issues with the tests validating and I think this was a byproduct of trying to get everything to work and ultimately I didn't even need it. I'll remove and update.

src/main/resources/com/ncc/aif/restricted_hypothesis_aif.shacl
sharknc: Same reasoning as above, I had issues with the test and was worried I was targeting more than just the importance values of other property nodes. I'll update this.

src/test/java/com/ncc/aif/ExamplesAndValidationTest.java
sharknc: ah okay, I was trying to make it look similar to the example. I'll remove the ldc: portion.

src/main/resources/com/ncc/aif/restricted_hypothesis_aif.shacl
sharknc: I don't know what is more efficient with respect to SPARQL. I can update it to UNION if you feel it may be a bit faster, just let me know.

ecurley-nextcentury: I really don't know. I was kinda hoping you could lend some insight from your query/SPARQL work with NIST. That being said, I think we can leave it the way it is. Perhaps the 'right way' will become evident when we finish our performance analysis.

src/test/java/com/ncc/aif/ExamplesAndValidationTest.java
sharknc: Updated.

src/test/java/com/ncc/aif/ExamplesAndValidationTest.java
sharknc: Updated.

PR #149, opened 2019-03-29
AIDA-608

ecurley-nextcentury:
Add AIFUtils.markInformativeJustification.
Modify shacl shapes for Entity/Relation/Event/Cluster.
Add ExamplesAndValidationTest.createEntityAndClusterWithInformativeJustification.


src/main/resources/com/ncc/aif/aida_ontology.shacl
```diff
@@ -133,10 +136,8 @@ aida:EventShape
     # events must have unique identifiers
     sh:nodeKind sh:IRI ;
 
-    # may have zero or more string labels for user consumption
-    sh:property [
-     sh:path skos:prefLabel ;
-     sh:datatype xsd:string ];
+    # may provide an informative justification
+    sh:property aida:InformativeJustificationPropertyShape ;
```
nextcen-dgemoets: Does the informative justification obviate the need for the prefLabel here and in RelationShape, or is it being removed for its own reason?  And if the latter, should it also be removed from EventArgumentShape and SentimentShape?  I do seem to recall you saying something about not supporting SKOS.

src/test/java/com/ncc/aif/ExamplesAndValidationTest.java
```diff
@@ -725,6 +725,31 @@ void createASimpleClusterWithHandle() {
...
+            final Resource justification = markTextJustification(model, typeAssertion, "HC00002Z0", 0, 10, system, 1d);
+            markInformativeJustification(vladimirPutin, justification);
+
+            final Resource putin = makeEntity(model, putinDocumentEntityUri, system);
+            markType(model, getAssertionUri(), putin, SeedlingOntology.Person, system, 1.0);
+            markName(putin, "Путин");
+
+            // create a cluster with prototype
+            final Resource putinCluster = makeClusterWithPrototype(model, getClusterUri(), vladimirPutin, vladName, system);
+            markInformativeJustification(putinCluster, justification);
```
nextcen-dgemoets: So is this saying that the same justification for the "vladimirPutin is a Person" is used as the justification that putin (with the Cyrillic name) is in the same-as cluster with vladimirPutin?  If so, I know that our examples don't have to be realistic, but should we consider creating a second justification for the cluster membership assertion?  Obviously it's "valid" either way.

src/main/resources/com/ncc/aif/aida_ontology.shacl
ecurley-nextcentury: It was removed because AIF doesn't use SKOS. I've been removing them as I come across them, but I'm OK to remove them all with this branch if that is preferable. Let me know what you'd like done.

nextcen-dgemoets: Sure, let's nuke 'em (and the prefix).  There are only two more.

src/test/java/com/ncc/aif/ExamplesAndValidationTest.java
ecurley-nextcentury: I'm content to leave it the way it is, but if you feel strongly that it should change, I can add more justifications

src/main/resources/com/ncc/aif/aida_ontology.shacl
ecurley-nextcentury: Removed. After merging with master, I was only able to find one more reference

PR #152, opened 2019-04-01
AIDA-685

craigwarsaw:
Add EdgeJustificationLimit.invalidZeroSpans
Add a few minor changes to make dumping output better by swapping comments

### DIFF
--- a/src/test/java/com/ncc/aif/ExamplesAndValidationTest.java
+++ b/src/test/java/com/ncc/aif/ExamplesAndValidationTest.java
@@ -1223,7 +1223,26 @@ public class ExamplesAndValidationTest {
                 testInvalid("NIST.invalid: edge justification contains at most two mentions");
             }
 
+
             @Test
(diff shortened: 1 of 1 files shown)


src/test/java/com/ncc/aif/ExamplesAndValidationTest.java
```diff
@@ -1223,7 +1223,26 @@ void invalid() {
                 testInvalid("NIST.invalid: edge justification contains at most two mentions");
             }
 
+
             @Test
+            void invalidZeroSpans() {
```
ecurley-nextcentury: This is tested in EdgeJustificationLimit.invalid(). If we want it to be in its own test, we should remove it from EdgeJustificationLimit.invalid().

craigwarsaw: Fixed. Removed zero span test from EdgeJustificationLimit.invalid()

PR #154, opened 2019-04-02
AIDA-703:  Check file size for TA3 output

nextcen-dgemoets:
Check file size for TA3 output, also refactored main method a bit.  I'm open to more refactoring, breaking up blocks of code into private methods.  Let me know.

To test, create a file about 5MB (it doesn't have to be a valid ttl) and invoke the validator (via  command line or IntelliJ) with command-line options of `--ldc -f <filename>.ttl --nist-ta3`
Tweak the file to be slightly more or less than 5MB and make sure the validator rejects the file that is > 5M.

Also, test some other combinations of valid and invalid kbs.  Note that --ldc now uses the LDC ontology, not Seedling.  But you can use an essentially empty file as a valid KB.

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,7 +92,7 @@ Ontology files can be found in `src/main/resources/com/ncc/aif/ontologies`:
 ### Validator return values
 Return values from the command-line validator are as follows:
 * `0 (Success)`.  There were no validation (or any other) errors.
-* `1 (Validation Error)`.	All specified files were validated but at least one failed validation.
+* `1 (Validation Error)`.	All specified files were validated but at least one failed validation.  Supersedes a File Error.
 * `2 (Usage Error)`.  There was a problem interpreting command-line arguments.  No validation was performed.
 * `3 (File Error)`.  A file was rejected, either due to an I/O error or because it didn't meet certain criteria.
   Logging indicates the nature of the problem(s).  Validation may have been performed on a subset of specified KBs.
--- a/README.md
+++ b/README.md
@@ -94,7 +94,9 @@ Return values from the command-line validator are as follows:
 * `0 (Success)`.  There were no validation (or any other) errors.
 * `1 (Validation Error)`.	All specified files were validated but at least one failed validation.
 * `2 (Usage Error)`.  There was a problem interpreting command-line arguments.  No validation was performed.
-* `3 (File Error)`.  There was a problem reading one or more files or directories.  Validation may have been performed on a subset of specified KBs.  If there is an error loading any ontologies or SHACL files, then no validation is performed.
+* `3 (File Error)`.  A file was rejected, either due to an I/O error or because it didn't meet certain criteria.
+  Logging indicates the nature of the problem(s).  Validation may have been performed on a subset of specified KBs.
+  If there is an error loading any ontologies or SHACL files, then no validation is performed.
 
 ### Running the validator in code
 To run the validator programmatically in Java code, first use one of the public `ValidateAIF.createXXX()`
--- a/src/main/java/com/ncc/aif/ValidateAIF.java
+++ b/src/main/java/com/ncc/aif/ValidateAIF.java
@@ -15,6 +15,9 @@ import org.apache.jena.util.FileUtils;
 import org.topbraid.shacl.validation.ValidationUtil;
 
 import java.io.File;
+import java.io.IOException;
+import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.text.DateFormat;
 import java.text.SimpleDateFormat;
 import java.util.*;
@@ -29,7 +32,9 @@ import java.util.*;
 public final class ValidateAIF {
 
     // Enum to track restrictions
-    public enum Restriction {NONE, NIST, NIST_HYPOTHESIS}
+    public enum Restriction {
+        NONE, NIST, NIST_HYPOTHESIS
+    }
 
     private static final String AIDA_SHACL_RESNAME = "com/ncc/aif/aida_ontology.shacl";
     private static final String NIST_SHACL_RESNAME = "com/ncc/aif/restricted_aif.shacl";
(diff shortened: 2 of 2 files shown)


src/main/java/com/ncc/aif/ValidateAIF.java
```diff
@@ -378,44 +385,89 @@ public static void main(String[] args) {
...
+                logger.info("No validation was performed.");
+            } else {
+                logger.info("The submitted KB was " + (invalidCount == 1 ? "invalid." : "valid."));
+            }
+        } else if (invalidCount > 0) { // Multiple file validation
+            final int validCount = fileCount - invalidCount - skipCount;
+            logger.info("Summary: " + validCount + " valid KB(s), " + invalidCount + " invalid KB(s)" +
+                    (skipCount > 0 ? ", " + skipCount + " skipped." : "."));
+        } else {
+            logger.info("All tested KBs were valid" + (skipCount > 0 ? " (" + skipCount + " skipped)." : "."));
```
ecurley-nextcentury: This seems overly complex. Why don't we have the single-file and multi-file outputs look the same? Couldn't we use the summary for every condition?

README.md
```diff
@@ -94,7 +94,9 @@ Return values from the command-line validator are as follows:
 * `0 (Success)`.  There were no validation (or any other) errors.
 * `1 (Validation Error)`.	All specified files were validated but at least one failed validation.
 * `2 (Usage Error)`.  There was a problem interpreting command-line arguments.  No validation was performed.
-* `3 (File Error)`.  There was a problem reading one or more files or directories.  Validation may have been performed on a subset of specified KBs.  If there is an error loading any ontologies or SHACL files, then no validation is performed.
+* `3 (File Error)`.  A file was rejected, either due to an I/O error or because it didn't meet certain criteria.
```
ecurley-nextcentury: Should we make these codes bit values (0, 1, 2, 4)? We could then return an error code that combines them and wouldn't have 1(Validation Error) clobber 3(File Error). What do you think?

src/main/java/com/ncc/aif/ValidateAIF.java
```diff
@@ -378,44 +385,89 @@ public static void main(String[] args) {
...
             final OntModel dataToBeValidated = ModelFactory.createOntologyModel();
-            dataToBeValidated.addLoadedImport(INTERCHANGE_URI);
-            dataToBeValidated.addLoadedImport(AIDA_DOMAIN_COMMON_URI);
-            try {
-                loadModel(dataToBeValidated, Files.asCharSource(fileToValidate, Charsets.UTF_8));
-            } catch (RuntimeException rte) {
-                logger.warn("---> Could not read " + fileToValidate + "; skipping.");
-                skipped = true;
-                skipCount++;
+            if (!skipped) {
```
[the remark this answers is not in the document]
nextcen-dgemoets: Yes, it's probably more readable...but I think there also might have been some other code in there earlier (maybe some logging?) as things were evolving that prevented such an assignment.  But now that the code is stable, I think your suggestion is an improvement-- thanks.

The funny thing is that after that change, we don't even actually need `skipped`.  The assignment above can just be used in the `if` check on the following line.  But IMO that's a little more readability to sacrifice just to save a boolean variable declaration and assignment.

src/main/java/com/ncc/aif/ValidateAIF.java
nextcen-dgemoets: The reason it's broken out this way is because it bothered me that you'd validate one KB (in this case a hypothesis) and it would say, "All KBs are valid (1 skipped)" if it was too big.  That's a misleading message.  Same thing if you validate one file and it can't be found or read. 
 Furthermore, if you're validating a directory of 10,000 KBs, just seeing, "at least one KB was invalid" isn't particularly helpful.  So I broke them into two cases (essentially `-f` and `-d`) and tracked the *number* of invalid KBs, not just if something had failed, which dated back to the Kotlin version of the validator.

So overall I think this change improves the user interface, at the cost of complexity-- which is why I broke it out into its own method.

src/main/java/com/ncc/aif/ValidateAIF.java
```diff
@@ -47,6 +52,7 @@
     private static Model nistModel;
     private static Model nistHypoModel;
     private static boolean initialized = false;
+    private static Logger logger;
```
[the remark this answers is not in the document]
nextcen-dgemoets: But the logger is used long before we've created the ValidateAIF object...it's only used in main() and the private methods called by main.  Perhaps I should declare it final and assign it in its static declaration?  I'm not completely sure how these loggers/LoggerFactories are initialized.

README.md
nextcen-dgemoets: We could but...that seems maybe a little unnecessarily complex.  It's true that Validation Error clobbers File Error, but the thinking was that it's more important and more common.  How about if I leave it as is but document this behavior here in the README?

ecurley-nextcentury: That's fair. The output also gives a good indication as to what's going on. I'm OK even if you don't want to update the README. I'll leave it up to you.

src/main/java/com/ncc/aif/ValidateAIF.java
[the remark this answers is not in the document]
ecurley-nextcentury: For some reason, I thought it was bound to the instance of the Validator. As it's only used in main and functions called from main, I think it's fine the way it is. Marking this resolved.

src/main/java/com/ncc/aif/ValidateAIF.java
nextcen-dgemoets: As we discussed in Teams, I improved summary output by including a report-style summary that looks the same whether you have single- or multiple-file validation.  I also included skipped non-ttl's in the skipCount to reflect better which files were skipped.

PR #155, opened 2019-04-03
Aida-694

jakevb:
Added python implementation of java code from AIDA-608.


python/aida_interchange/aifutils.py
```diff
@@ -404,6 +404,13 @@ def mark_importance(g, resource, importance):
     g.add((resource, AIDA_ANNOTATION.importance, Literal(importance, datatype=XSD.int)))
 
 
+def mark_informative_justification(g, resource, informative_justification):
+    """
+    Mark resource as having an informativeJustification value
```
bdpham: Add documentation for parameters

PR #156, opened 2019-04-03
Aida 674

sharknc:
- Created hypothesis restriction to enforce all clusters must be homogeneous by base class (Entity, Event, or Relation)
- Created valid and invalid test cases to validate the new aforementioned hypothesis restriction
- Updated all other hypothesis test cases that were impacted by the required importance value
- Updated all Hypothesis examples to include more detailed notification strings to help disambiguate output for https://nextcentury.atlassian.net/wiki/spaces/AIDAC/pages/507347434/Example+AIF+that+pass+and+fail+NIST+validations

### DIFF
--- a/src/main/resources/com/ncc/aif/restricted_hypothesis_aif.shacl
+++ b/src/main/resources/com/ncc/aif/restricted_hypothesis_aif.shacl
@@ -148,8 +148,7 @@ aida:HypothesisClusterMembersSameAsBaseClass
 aida:HypothesisClusterMembersShape
     a sh:NodeShape ;
     sh:targetClass aida:ClusterMembership ;
-    sh:sparql aida:HypothesisClusterMembersSameAsBaseClass
-    .
+    sh:sparql aida:HypothesisClusterMembersSameAsBaseClass .
 
 ########################
 # 4.3 TA3 #1.a All KEs in model should be referenced by hypothesis
--- a/src/main/resources/com/ncc/aif/restricted_hypothesis_aif.shacl
+++ b/src/main/resources/com/ncc/aif/restricted_hypothesis_aif.shacl
@@ -128,19 +128,20 @@ aida:HypothesisEntityClusterRequireHandle
 #------------------------
 aida:HypothesisClusterMembersSameAsBaseClass
     a sh:SPARQLConstraint ;
-    sh:message "All Clusters must be homogeneous by base class (Entity, Event, or Relation). Found {?value}" ;
+    sh:message "All Clusters must be homogeneous by base class (Entity, Event, or Relation). Found {?prototype} in cluster {?cluster}" ;
     sh:select """
         PREFIX rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#>
         PREFIX aida:  <https://tac.nist.gov/tracks/SM-KBP/2018/ontologies/InterchangeOntology#>
-        SELECT $this (COUNT(?cluster) AS ?value)
+        SELECT $this ?cluster ?prototype  (COUNT(?cluster) AS ?value)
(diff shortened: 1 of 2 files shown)


src/main/resources/com/ncc/aif/restricted_hypothesis_aif.shacl
```diff
@@ -123,6 +123,33 @@ aida:HypothesisEntityClusterRequireHandle
...
+########################
+# 4.3 TA3 #7  Clusters must be homogeneous by base class (Entity, Event, or Relation)
+#------------------------
+aida:HypothesisClusterMembersSameAsBaseClass
+    a sh:SPARQLConstraint ;
+    sh:message "All Clusters must be homogeneous by base class (Entity, Event, or Relation). Found {?value}" ;
+    sh:select """
+        PREFIX rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#>
+        PREFIX aida:  <https://tac.nist.gov/tracks/SM-KBP/2018/ontologies/InterchangeOntology#>
+        SELECT $this (COUNT(?cluster) AS ?value)
```
ecurley-nextcentury: Can't tell from the return value which cluster or Entity/Event/Relation violates this. Rather than a count, this should probably indicate which Entity/Event/Relation doesn't match which cluster.

src/test/java/com/ncc/aif/ExamplesAndValidationTest.java
```diff
@@ -1821,8 +1824,65 @@ void validEventRelationAndEventRelationEdge() {
...
+                final Resource relation = makeRelation(model, getUri("relation-1"), system);
+                final Resource relationCluster = makeClusterWithPrototype(model, getClusterUri(), relation,
+                        "Relation", system);
+
+                markImportance(relationCluster, 103);
+                markJustification(addType(relation, SeedlingOntology.GeneralAffiliation_APORA), justification);
+
+                final Resource relationEdge = markAsArgument(model, relation, SeedlingOntology.GeneralAffiliation_APORA_Affiliate,
+                        entity, system, 1d, getAssertionUri());
+                markImportance(relationEdge, 102);
```
ecurley-nextcentury: Everything from here, up is duplicated in valid(). This should be moved into a setup() function to only have the code live in one place.

sharknc: Updated.

src/main/resources/com/ncc/aif/restricted_hypothesis_aif.shacl
sharknc: Updated.

ecurley-nextcentury:
Your shacl works, but having the values in the message is a little different than what I've seen. If you want to put the violation into report parameters, take a look at this:
[674Suggestion.patch.txt](https://github.com/NextCenturyCorporation/AIDA-Interchange-Format/files/3041259/674Suggestion.patch.txt)

I tried to add this to our conversation, but it wouldn't let me attach files there.

sharknc:
> Your shacl works, but having the values in the message is a little different than what I've seen. If you want to put the violation into report parameters, take a look at this:
> [674Suggestion.patch.txt](https://github.com/NextCenturyCorporation/AIDA-Interchange-Format/files/3041259/674Suggestion.patch.txt)
> 
> I tried to add this to our conversation, but it wouldn't let me attach files there.

I see. I thought the goal was to get the error in the resultMessage, but after seeing it in the report seems like a much better solution. Also, I didn't realize there were other pre-bound variables like `$value` that we can attach to, very cool, thanks for the heads up on that. I have made your suggested updates so the report now shows the erroneous cluster and cluster member in the report.

PR #157, opened 2019-04-03
Upgrade to Kotlin 1.3

rgabbard:
This fixes a deprecated Kotlin standard library dependency (used by the ColdStart converter)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -10,7 +10,7 @@
 
 
     <properties>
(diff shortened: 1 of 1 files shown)


rgabbard:
@nextcen-dgemoets : I think the newer version of Kotlin expects IntelliJ 2019.1, but the mismatch shouldn't affect anything for other users. Thanks for the merge!

PR #159, opened 2019-04-08
Aida 701

jakevb:
Modified shacl file to allow KB link added to clusters.
Added example case for cluster KB link into ExamplesAndValidationTest.


src/main/resources/com/ncc/aif/aida_ontology.shacl
```diff
@@ -1067,6 +1067,9 @@ aida:ClusterShape
    # may bear system-private data
    sh:property aida:PrivateDataPropertyShape ;
 
+   # my provide a link
+   sh:property aida:LinkPropertyShape ;
```
nextcen-dgemoets: AIDA-701 states 

> linkTarget must not be an empty string, but no further validation on the value is required.

I wonder if the `aida:LinkPropertyShape` should be enhanced to include this check.  I've started a Teams conversation on the topic, so we'll see.  If we do end up checking for an empty linkTarget, then we should probably have an InvalidTest that creates an empty link.

src/test/java/com/ncc/aif/ExamplesAndValidationTest.java
```diff
@@ -1035,6 +1035,23 @@ void createEntityWithDiskBaseModelAndWriteOut() {
...
+            Resource typeAssertionJustification = makeTextJustification(model, "NYT_ENG_20181231",
+                    42, 143, system, 0.973);
+            Resource newEntity = makeEntity(model, getEntityUri(), system);
+            markJustification(addType(newEntity, SeedlingOntology.Person), typeAssertionJustification);
+            Resource cluster = makeClusterWithPrototype(model, getClusterUri(), newEntity, "handle", system);
+
+            // link cluster to a KB
+            linkToExternalKB(model, cluster, "freebase:FOO", system, .398);
+
+            testValid("Cluster with Link and confidence is valid.");
```
nextcen-dgemoets: To be in line with other `testValid` calls, how about, "Create a cluster with link and confidence."

src/test/java/com/ncc/aif/ExamplesAndValidationTest.java
```diff
@@ -1035,6 +1035,23 @@ void createEntityWithDiskBaseModelAndWriteOut() {
...
+        /**
+         * A cluster should be able to contain a link to one or more external KBs
+         */
+        @Test
+        void createClusterWithLinkAndConfidence() {
+            // create a valid entity and cluster
+            Resource typeAssertionJustification = makeTextJustification(model, "NYT_ENG_20181231",
+                    42, 143, system, 0.973);
+            Resource newEntity = makeEntity(model, getEntityUri(), system);
+            markJustification(addType(newEntity, SeedlingOntology.Person), typeAssertionJustification);
```
nextcen-dgemoets: It looks like this code was borrowed from the NIST examples (setup?) where we use `addType()` and justify all type assertions.  Although it's not bad or wrong to do any of this, it is a bit non-parallel with other non-NIST AIF tests, and it will look more so once we break this file up into AIF, NIST, and NIST-hypothesis in AIDA-719.  This will also mean the Example on the Confluence page (if we add it) will be streamlined.

So I'd suggest changing the example a bit to be more like the other AIF examples above it.  Take a look at `createASimpleCluster()` where it creates an entity, marks its type, and eventually adds it to a cluster.  In fact, one could argue that the code could be the same as `createASimpleCluster()`, but with the `linkToExternalKB()` call added.

src/main/resources/com/ncc/aif/aida_ontology.shacl
nextcen-dgemoets: Okay, based on offline conversation, it looks like nothing should be added to AIF other than the link itself, but as the NIST restriction is finalized, perhaps we'll add a check for empty to restricted AIF.

PR #161, opened 2019-04-09
Aida 708

bdpham:
Created a script to validate the examples from Examples.py

### DIFF
--- a/python/tests/Examples.py
+++ b/python/tests/Examples.py
@@ -17,8 +17,6 @@ class Examples(unittest.TestCase):
     test_dir_path = os.environ.get("DIR_PATH", None)
     def new_file(self, g, test_name):
         if self.test_dir_path is not None:
-            if not os.path.exists(self.test_dir_path):
-                os.makedirs(self.test_dir_path)
             f = open(self.test_dir_path + "/" + test_name, "wb+")
             f.write(g.serialize(format='turtle'))
             f.close
@@ -678,11 +676,11 @@ class Examples(unittest.TestCase):
         aifutils.mark_as_possible_cluster_member(g, president_cluster, trump_cluster, .6, system)
 
         # write graph to file
-        self.new_file(g, "../test_read_and_write.ttl")
+        self.new_file(g, "test_read_and_write.ttl")
 
         # create new graph and read in file
         graph = Graph()
-        graph.parse("../test_read_and_write.ttl", format='turtle')
+        graph.parse("test_read_and_write.ttl", format='turtle')
 
         # verify that one of the trump entities exists in new graph object
         dtrump = URIRef("http://www.test.edu/entities/4")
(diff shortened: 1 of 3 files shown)


python/tests/Examples.py
```diff
@@ -13,6 +14,15 @@
 # Running these tests will output the examples to the console
 class Examples(unittest.TestCase):
 
+    test_dir_path = os.environ.get("DIR_PATH", None)
+    def new_file(self, g, test_name):
```
sharknc: You should move this into the pyValidator.sh script. This script should handle both creating and deleting this directory.

python/tests/Examples.py
```diff
@@ -13,6 +14,15 @@
 # Running these tests will output the examples to the console
 class Examples(unittest.TestCase):
 
+    test_dir_path = os.environ.get("DIR_PATH", None)
```
sharknc: I would move setting this directory variable down in the main method below. As you mentioned, you could then check to see if the path exists and warn the user if the directory has not been created yet.

python/tests/pyValidator.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+
+set -x
```
sharknc: Is there something in the debug info that the user should see when running this script? If not, I would recommend removing this flag.

python/tests/pyValidator.sh
```diff
@@ -0,0 +1,23 @@
...
+
+# run Examples.py
+# DIR_PATH is the directory where example files are to be written
+DIR_PATH=$test_directory python3 Examples.py
+
+# validate test file directory
+../../target/appassembler/bin/validateAIF --ont $ontology_dir -d $test_directory
+
+# if all files are valid, delete the test file directory
+rm -r $test_directory
```
sharknc: It looks like we are relying on the fact that the validator will throw an exception when validation fails which will force a failure of this script and no ultimately never call this `rm -r` command. Is that correct? I am not closely tracking all the changes @nextcen-dgemoets is currently making to the AIF validator, but I believe he is updating how failures are handled. It may be worth making sure that his changes won't inadvertently break the functionality of this script.

bdpham: I will sync up with him today about this. All other requested changes have been made.

nextcen-dgemoets: Typically, the validator *won't* throw an Exception and terminate abnormally.  The [Validator Return Values](https://github.com/NextCenturyCorporation/AIDA-Interchange-Format/#validator-return-values) section of the AIF README describes the return codes.  If you only want to delete test directories when validation is successful, then look for a return code of zero.

Also note that in [PR#165](https://github.com/NextCenturyCorporation/AIDA-Interchange-Format/pull/165) for [AIDA-711](https://nextcentury.atlassian.net/browse/AIDA-711), I am adding a `-o` switch for sending validation output to a file.  This may or may not help you here, and of course it's not quite in master yet.  Maybe by the end of the day?

PR #162, opened 2019-04-09
AIDA-688:  Update ScalingTest to generate valid/invalid LDC restricted AIF

nextcen-dgemoets:
Updated the ScalingTest to use LDC Ontology and generate valid or invalid restricted AIF.

To test, generate valid and invalid files and validate with the AIF validator (saving invalid results to a file).  Inspect the report from the invalid file and ensure that types and numbers of violations are in line with expectations.  Note that validation time has increased significantly with the latest NIST restrictions.  On my laptop, validating a model with 1000 entities and 300 events took 26 minutes.

I'm hoping that as part of the review process, we can come up with better/reasonable frequency constants (FREQ_*) for each of the restrictions.  I have set them all to a "default" value of 0.3.  Note that this number doesn't necessarily mean "30% of the time, the given restriction will be violated," as some of the restrictions interrelate.

Note that AIDA-688 talks about generating models of "significant complexity," which basically means, "models that make sense."  That part of the issue has been removed from this issue for several reasons:
- This ticket was time-boxed to 5 story points
- As part of AIDA-553, I will be creating a large "model that makes sense" generated from the LDC Annotations (multiplied by 1000 or so)
- Generating a model that resembles that of a researcher isn't really in my wheelhouse

Once a new model is created in AIDA-553 from the LDC Annotations, we can decide if that model satisfies the need originally stated in AIDA-688 or if we need to create a new ticket.

### DIFF
--- a/src/test/java/com/ncc/aif/ScalingTest.java
+++ b/src/test/java/com/ncc/aif/ScalingTest.java
@@ -2,8 +2,8 @@ package com.ncc.aif;
 
 import ch.qos.logback.classic.Level;
 import ch.qos.logback.classic.Logger;
+import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
-import com.google.common.io.Resources;
 import org.apache.jena.ext.com.google.common.collect.ImmutableList;
 import org.apache.jena.query.Dataset;
 import org.apache.jena.rdf.model.Model;
@@ -13,12 +13,8 @@ import org.apache.jena.rdf.model.StmtIterator;
 import org.apache.jena.riot.RDFDataMgr;
 import org.apache.jena.riot.RDFFormat;
 import org.apache.jena.tdb.TDBFactory;
-import org.apache.jena.vocabulary.RDF;
-import org.apache.jena.vocabulary.SKOS;
-import org.apache.jena.vocabulary.XSD;
 
 import java.io.File;
-import java.nio.charset.StandardCharsets;
 import java.nio.file.*;
 import java.util.*;
 
@@ -26,7 +22,7 @@ import static com.ncc.aif.AIFUtils.*;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
 /**
- * Test to see how large we can scale AIF with the LDC Seedling ontology.  AIF uses a Jena-based model,
+ * Test to see how large we can scale AIF with the LDC ontology (LO).  AIF uses a Jena-based model,
  * and so we rely on that to determine how large it can get.  Change whether you are using a memory based model
  * or a disk-based model (TDB) below in the line that defines modelTypeToUse.  MEMORY is faster, but
  * is more limited;  TDB is slower but can handle more statements.
@@ -54,6 +50,9 @@ import static org.junit.jupiter.api.Assertions.assertTrue;
  * The smaller ones are difficult to read, the large ones do not use prefixes. turtle_pretty is readable and not too large.
  * By default, this class runs a scaling test using the Turtle Pretty output type.
  * <p>
+ * This class can generate either valid or invalid restricted AIF output, as specified by the -i switch.  Invalid output
+ * can be configured by tweaking the FREQ_* class constants.  The higher the number, the more often that particular
+ * restriction will be violated by the generated TTL file.
  * Run with:
  * %  mvn exec:java -Dexec.mainClass="com.ncc.aif.ScalingTest" -Dexec.classpathScope="test" -Dexec.args="[arguments]"
  * where arguments are:
@@ -62,6 +61,7 @@ import static org.junit.jupiter.api.Assertions.assertTrue;
  *       -o   run a single test on different output types (incompatible with -s and -v)
  *       -t   use tdb model (default is to use in-memory)
  *       -v   do validation (default is to not do validation, incompatible with -o)
+ *       -i   generate invalid output (default is valid output)
  * </pre>
  */
 public class ScalingTest {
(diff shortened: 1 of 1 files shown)


src/test/java/com/ncc/aif/ScalingTest.java
```diff
@@ -77,20 +77,63 @@
...
+        PreventShotVideo,               // Video must use aida:KeyFrameVideoJustification. Remove ShotVideoJustification
+        FlatClusters,                   // Members of clusters are entity objects, relation objects, and event objects (not clusters)
+        EverythingClustered,            // Entity, Relation, and Event object is required to be part of at least one cluster.
+        ConfidenceValueRange,           // Each confidence value must be between 0 and 1
+        ClusterHasIRI,                  // Entity, Relation, and Event clusters must have IRI
+        JustifyTypeAssertions,          // Each entity/relation/event type statement must have at least one justification
+        NameMaxLength                   // Each entity/filler name string is limited to 256 UTF-8 characters
+    }
+
+    // Tweak these values to introduce more/fewer invalid content in the generated KB.
```
craigwarsaw: I don't have anything against these particular values.  However, I assume this is intended to be a percentage, but the comment isn't clear.  It should indicate what the range of values should be and whether a higher value introduces more invalid content.

src/test/java/com/ncc/aif/ScalingTest.java
```diff
@@ -259,62 +326,114 @@ private void addEventOrRelation() {
...
+                    ImmutableSet.of(resource), ImmutableSet.of(justification), system,
+                    preserveValidity(NIST_RESTRICTION.ConfidenceValueRange) ? 1.0 : -1.0);
+        }
+        markJustification(resource, compound);
+
+        // Add some private data
+        markPrivateData(model, resource, "{ 'provenance' : '" + docId + "' }", system);
+    }
+
+    // Returns true when invalid content should be generated
```
craigwarsaw: These two methods are confusing.
Why is the comment on checkChaos identical to the comment on preserveValidity when they are doing opposite things (I think).
To me, the method name "checkChaos" doesn't imply the opposite of preserveValidity.

src/test/java/com/ncc/aif/ScalingTest.java
nextcen-dgemoets: Good idea-- improved comment.

src/test/java/com/ncc/aif/ScalingTest.java
nextcen-dgemoets: The comment was definitely a copy/paste error-- fixed.
I changed checkChaos to introduceChaos.  I'm open to suggestions.  `introduceInvalidity()`?  `introduceInvalidContent()`?

nextcen-dgemoets:
> The generated output does [sic] use the ldcOnt: prefix (I used the -s flag). For example, it generates:
> rdf:object <https://tac.nist.gov/tracks/SM-KBP/2018/ontologies/LDCOntology#Person>

This is just what happens when I use the constants in com.ncc.aif.LDCOntology.  Is there something I should be doing differently here?

> Although not implemented as a restriction yet, this will soon generate all invalid output because the sourceDocument attribute isn't added.

Yep, unfortunately this class will need to be maintained as restrictions change.

src/test/java/com/ncc/aif/ScalingTest.java
craigwarsaw: introduceChaos is much better

src/test/java/com/ncc/aif/ScalingTest.java
```diff
@@ -77,20 +77,64 @@
...
+    private String filename = "scalingdata-" + entityCount + "x" + eventCount + ".ttl";
 
     // Whether to use in memory or disk based.
     // Note:  TDB2 requires transactions, which we do not do!  Do not use it!
     private enum MODEL_TYPE {
         MEMORY, TDB
     }
 
+    // These NIST restrictions correspond to (and should be updated from) com/ncc/aif/ExamplesAndValidationTest.java.
+    private enum NIST_RESTRICTION {
```
ecurley-nextcentury: It seems to me like you're building an enum with values for the frequency for violating restrictions. Would it make sense to just include these values in the enum? https://docs.oracle.com/javase/tutorial/java/javaOO/enum.html

src/test/java/com/ncc/aif/ScalingTest.java
```diff
@@ -259,62 +327,114 @@ private void addEventOrRelation() {
...
 
         addJustificationAndPrivateData(typeAssertion);
 
         // Make two arguments
         Resource argument = markAsArgument(model, eventResource,
-                SeedlingOntology.Business_DeclareBankruptcy_Place,
+                LDCOntology.Conflict_Attack_AirstrikeMissileStrike_Place,
                 getRandomEntity(), system, 0.785, getAssertionUri());
-        addJustificationAndPrivateData(argument);
+        if (preserveValidity(NIST_RESTRICTION.RestrictJustification_Edges)) {
```
ecurley-nextcentury: Why do you apply some restrictions to only Entities/Events/Relations, but you apply this to both Events and Relations?

src/test/java/com/ncc/aif/ScalingTest.java
```diff
@@ -259,62 +327,114 @@ private void addEventOrRelation() {
...
-        Resource typeResource = SeedlingOntology.Business_DeclareBankruptcy;
+        Resource typeResource = LDCOntology.Conflict_Attack_AirstrikeMissileStrike;
         Resource typeAssertion = markType(model, getAssertionUri(), eventResource, typeResource, system, 1.0);
 
         addJustificationAndPrivateData(typeAssertion);
 
         // Make two arguments
         Resource argument = markAsArgument(model, eventResource,
-                SeedlingOntology.Business_DeclareBankruptcy_Place,
+                LDCOntology.Conflict_Attack_AirstrikeMissileStrike_Place,
```
ecurley-nextcentury: As we only have people, results, and numbers in our model, we might want to find events/relations that can have those things as arguments. It might make it easier to actually separate those things into their own collections so we can assign arguments correctly, but we don't have to do that now. The model doesn't really have to "make sense", but we should try to avoid violating the Ontology if we can.

src/test/java/com/ncc/aif/ScalingTest.java
```diff
@@ -259,62 +327,114 @@ private void addEventOrRelation() {
...
+                    preserveValidity(NIST_RESTRICTION.ConfidenceValueRange) ? 1.0 : -1.0);
+        }
+        markJustification(resource, compound);
+
+        // Add some private data
+        markPrivateData(model, resource, "{ 'provenance' : '" + docId + "' }", system);
+    }
+
+    // Returns true when invalid content should be generated
+    private boolean introduceChaos(NIST_RESTRICTION restriction) {
```
ecurley-nextcentury: Why not just use !preserveValidity()?

src/test/java/com/ncc/aif/ScalingTest.java
nextcen-dgemoets: Yes-- I would love any suggestions on what ontology types we should use and where.  I tried to find something in LDC similar/analogous to the Seedling version that was already there, but once I got it to generate valid AIF, I stopped there.  But I'd love to improve it.

src/test/java/com/ncc/aif/ScalingTest.java
nextcen-dgemoets: Because it's easier to read.  :)

ecurley-nextcentury: Haha, if you say so.

src/test/java/com/ncc/aif/ScalingTest.java
ecurley-nextcentury: LDCOwlOntology lists out types that each argument for relations/events can have. I just did a search for ldcOnt:Person and there are a couple (like Evaluate.Sentiment) where a person can be the "holder" and the "target".  Conflict.Attack can also have a person for "attacker" and "target". Maybe we just need to change which argument we're filling for the attack event?

src/test/java/com/ncc/aif/ScalingTest.java
nextcen-dgemoets: My approach was to find a reasonable place or places to introduce chaos (invalidities) in the generated model and apply the restrictions when I could (if it didn't interfere with another restriction), with the goal to generate enough of those invalidities on average.  So don't bury a restriction only in an edge case.

Because there tend to be fewer events and relations than entities, I decided to apply that particular restriction to both events and relations-- to make sure that enough invalidities were generated to make it meaningful.  That's also why EverythingClustered and NameMaxLength can only be introduced for entities, because there are many more of them.

Not sure if this makes sense, but let me know if you think a given restriction should be added/removed from one place or another.  If you think one restriction is being generated too much or too little-- with all of the FREQ constants currently set to 0.3-- then let me know.

src/test/java/com/ncc/aif/ScalingTest.java
nextcen-dgemoets: In the Seedling version of ScalingTest, Business_DeclareBankruptcy was being filled by Business_DeclareBankruptcy_Place and Business_DeclareBankruptcy_Organization (basically, a place and a "person."  I tried to do something analogous with Conflict_Attack_AirstrikeMissileStrike being filled by Conflict_Attack_AirstrikeMissileStrike_Place and Conflict_Attack_AirstrikeMissileStrike_Attacker-- also a place and a "person."  They were both under ldcOnt:Conflict.Attack and both were event argument types.  That was the extent of my investigation (and that the AIF was valid).  How should we change the attack event fillers?

nextcen-dgemoets: Made the changes we discussed on the phone so that things are more ontologically valid.

src/test/java/com/ncc/aif/ScalingTest.java
nextcen-dgemoets: Cool, didn't know enums could do that.  I enhanced the Enum and removed the constants and map.

nextcen-dgemoets:
```
The generated output does [sic] use the ldcOnt: prefix (I used the -s flag). For example, it generates:
rdf:object https://tac.nist.gov/tracks/SM-KBP/2018/ontologies/LDCOntology#Person

```
Fixed the constant used by the prefix.  It was missing the `#`.

PR #164, opened 2019-04-12
Aida 716

sharknc:
- Created NIST restriction to enforce all justifications include a source and sourceDocument
- Created valid and invalid test cases to validate the new aforementioned NIST restriction
- Updated all other NISTExamples and Hypothesis test cases that were impacted by the required source and sourceDocument data in the justifications.

### DIFF
--- a/src/main/resources/com/ncc/aif/restricted_aif.shacl
+++ b/src/main/resources/com/ncc/aif/restricted_aif.shacl
@@ -223,7 +223,7 @@ aida:JustificationSourceDocumentShape
     sh:property [
         sh:path aida:sourceDocument ;
         sh:minCount 1 ;
-        sh:message "Justifications require source and a sourceDocument" ;
+        sh:message "Justifications require a sourceDocument" ;
     ]
     .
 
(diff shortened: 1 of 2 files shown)


src/main/resources/com/ncc/aif/restricted_aif.shacl
```diff
@@ -213,6 +213,20 @@ aida:RequiredJustificationPropertyShape
...
+# Justifications require a source document and a source
+# The source restriction is defined on the aida:SharedJustificationShape in the aida_ontology.shacl
+#------------------------
+aida:JustificationSourceDocumentShape
+    a sh:NodeShape ;
+    sh:targetClass aida:Justification ;
+    sh:property [
+        sh:path aida:sourceDocument ;
+        sh:minCount 1 ;
+        sh:message "Justifications require source and a sourceDocument" ;
```
ecurley-nextcentury: We should make this explicit for sourceDocument

sharknc: fixed.

nextcen-dgemoets:
Merging this PR has caused VERDI tests to fail, because the models created by LDCModelGeneratorTest no longer validate because their justifications lack a sourceDocument.

This also affects the models created by the ScalingTest, but there is no test for that.

PR #167, opened 2019-04-17
Aida 730

sharknc:
- Created initial Dockerfile to build AIF validator container
- Created build and run scripts to help user build the container and run the container
- Created initial README.md describing how to build and run the container
- We are still deciding how this Docker container will be used in an AWS Batch pipeline. Because there are so many unknowns at this point, the docker container runs indefinitely in order to allow the reviewer to enter into the container and verify that all the AIF validation has been set up properly.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -31,7 +31,7 @@ RUN \
   apt-get install -y maven
 
 #-------------------------------------------------------------
-# Install SPARQL Evaluation Tool
+# Install AIF Interchange Format
 #------------------------------------------------------------- 
 # Create directory structure
 ENV VALIDATION_HOME /opt/aif-validator
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,0 +1,63 @@
+FROM ubuntu:18.04
+MAINTAINER patrick.sharkey@nextcentury.com
+
+# Prevent dialog during apt install
+ENV DEBIAN_FRONTEND noninteractive
+
+# Update system packages and remove any downloaded files
+RUN \
+  rm -rf /var/lib/apt/lists/* \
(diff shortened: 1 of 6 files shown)


docker/Dockerfile
```diff
@@ -0,0 +1,63 @@
...
+ENV JAVA_HOME /usr/lib/jvm/java-8-openjdk-amd64
+
+#-------------------------------------------------------------
+#  Install Maven
+#-------------------------------------------------------------
+RUN \
+  apt-get install -y maven
+
+#-------------------------------------------------------------
+# Install SPARQL Evaluation Tool
```
nextcen-dgemoets: AIF Validator (or AIF repository)

docker/run.sh.example
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+docker run --rm -d \
+	--name aif-validator \
+    	-it aif-validator
```
nextcen-dgemoets: Maybe this is just part of the TBDs in this, or it's because I've never used Docker, but does this actually run the validator?  I would think that it would need to invoke the `target/appassembler/bin/validateAIF` script, which invokes the `ValidateAIF` class file.

docker/Dockerfile
sharknc: Good catch, I took the base docker from the SPARQL eval tool. I must have missed updating this.

docker/run.sh.example
sharknc: Yep, this is part of the TBD. We first need to decide how this docker container is going to be used. Because we have not decided this yet I made the docker container not do anything except tail continuously so  the reviewer can go into the container to review the installation of the validator. I mention this in the README at the very end.

PR #168, opened 2019-04-18
AIDA-604

sharknc:
- Updated base AIF rule to require only one justification in compound justification.
- Created single test to validate that a single justification in a compound justification is valid

### DIFF
--- a/src/main/resources/com/ncc/aif/aida_ontology.shacl
+++ b/src/main/resources/com/ncc/aif/aida_ontology.shacl
@@ -779,7 +779,7 @@ aida:CompoundJustificationMinimum
    a sh:PropertyShape ;
    sh:path aida:containedJustification ;
    sh:minCount 1 ;
-   sh:message "aida:CompoundJustification requires at least 1 justifications" .
+   sh:message "aida:CompoundJustification requires at least 1 justification" .
 
 aida:CompoundJustificationTypes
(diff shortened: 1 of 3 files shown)


nextcen-dgemoets:
Actually, do we still need the following in restricted_aif.shacl?

[# Turn off CompoundJustificationMinimum globally
aida:CompoundJustificationMinimum sh:deactivated true .
](url)

sharknc:
> @nextcen-dgemoets made a good observation - try taking out the deactivation of the CompoundJustificationMinimum to see if that still works as expected.
> I would like to see another RestrictJustification invalid test with one CompoundJustification with 0 justifications and another CompoundJustification with 3 justifications (or maybe augment the existing invalid test to show those cases).

Okay I will implement those.

sharknc:
> > @nextcen-dgemoets made a good observation - try taking out the deactivation of the CompoundJustificationMinimum to see if that still works as expected.
> > I would like to see another RestrictJustification invalid test with one CompoundJustification with 0 justifications and another CompoundJustification with 3 justifications (or maybe augment the existing invalid test to show those cases).
> 
> Okay I will implement those.

I have updated the RestrictJustification tests to include two invalid tests made up of a CompoundJustification with zero justifications and a CompoundJustification with three justifications.

PR #169, opened 2019-04-19
Aida 757

sharknc:
- Created NIST restriction to enforce that all informative mentions on a KE points to a different sourceDocument
- Created valid and invalid test cases to validate the new aforementioned NIST restriction
- Created valid and invalid test cases to valid that informative mentions can only live on Clusters, Entities, Relations, and Events.

### DIFF
--- a/src/main/resources/com/ncc/aif/restricted_aif.shacl
+++ b/src/main/resources/com/ncc/aif/restricted_aif.shacl
@@ -248,7 +248,7 @@ aida:InformativeJustificationMembersUniqueParentDoc
       	    {
       	        SELECT $this (COUNT(DISTINCT ?parentDoc) as ?pdCount) (COUNT(DISTINCT ?mentions) as ?mCount)
       	        WHERE {
-      	            ?statement aida:informativeJustification ?mentions .
+      	            $this aida:informativeJustification ?mentions .
       	            ?mentions aida:sourceDocument ?parentDoc
       	        }
       	        GROUP BY $this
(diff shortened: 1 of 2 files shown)


src/main/resources/com/ncc/aif/restricted_aif.shacl
```diff
@@ -231,4 +231,31 @@ aida:JustificationSourceDocumentShape
...
+      	WHERE {
+      	    $this aida:informativeJustification ?mentions .
+      	    ?mentions aida:sourceDocument ?parentDoc
+      	}
+      	GROUP BY $this
+      	HAVING (COUNT(DISTINCT ?parentDoc) < COUNT(DISTINCT ?mentions))
+    """
+    .
+
+aida:InformativeJustificationMembersShape
```
nextcen-dgemoets: `aida:InformativeJustificationMembersShape` doesn't appear to be used in any of the SHACL files.  Can it be removed?

src/test/java/com/ncc/aif/ExamplesAndValidationTest.java
```diff
@@ -1604,6 +1604,123 @@ void valid() {
             }
         }
 
+        @Nested
+        class InformativeJustification {
+
+            @Test
+            void invalidArgumentInformativeJustification() {
```
nextcen-dgemoets: The only thing I'd like to note is that this doesn't fail because of the recent changes (e.g., because of `aida:InformativeJustificationMembersUniqueParentDoc`).  It fails because `aida:RelationArgumentShape` is a closed shape.  If the the shape is opened (via editing aida_ontology.shacl) it passes validation.

Maybe that's perfectly okay...I just thought I'd point it out.

sharknc: This is a fair point. I'll remove this test.

sharknc: Fixed.

src/main/resources/com/ncc/aif/restricted_aif.shacl
sharknc: This is needed in order to get the SPARQL constraint to trigger.

craigwarsaw:
FWIW - I generated the TTL from the examples and sent them to Shahzad. He reviewed and confirmed that we are good.  So we can say that we have implemented the restriction as NIST intended.

PR #174, opened 2019-04-24
AIDA-676: Add subClassOf correctly to LDC ontologies

ecurley-nextcentury:


ecurley-nextcentury:
> I don't see a AIDA-676 PR for the VERDI code that generates this, so I'm just looking at the output (LDCOntology).

The VERDI portion is based off 748 so I won't create a PR until 748 is merged.
> * Is there no SID.Political in the source data?

There is not
> * Why did you remove the rdfs:label to the Relation Types?

https://www.w3.org/TR/rdf-schema/#ch_label. As it was the same as the actual name, there was no point in keeping it.
>   I'm sure these last two bullets are due to some deficiency in the source data, but can you characterize what is going on so that I can report it to LDC/NIST?

As with SID.Political none of these other super-classes exist. I not sure it's a deficiency, but it is certainly what LDC encoded in their ontology.

PR #179, opened 2019-05-01
Aida 718: Develop AIF test utilities (also includes Aida-719)

nextcen-dgemoets:
Work performed:
- Created TestUtils class for base AIF test functionality and tasks common to all tests;
- Created NistTestUtils subclass for restricted AIF tests;
- Created NistHypothesisUtiils subclass for restricted AIF TA3 tests;
- Broke ExamplesAndValidationTests into three files (AIDA-719) which all run as part of an install;
- Updated all tests to use TestUtils class (and subclasses), except for base AIF tests that serve as excellent code samples with comments.

The most sure-fire (albeit tedious) way to test this is by comparing the output of the tests in the master branch to the output created by these tests.  I have run valid and invalid tests on areas that I changed the most to ensure they fail for the right reasons.  Of course, you'll want to refer to the original [ExamplesAndValidationTests](https://github.com/NextCenturyCorporation/AIDA-Interchange-Format/blob/master/src/test/java/com/ncc/aif/ExamplesAndValidationTest.java).


src/test/java/com/ncc/aif/TestUtils.java
```diff
@@ -0,0 +1,466 @@
...
+     * @return the created type assertion resource
+     */
+    Resource addType(Resource resource, Resource type, double confidence) {
+        return markType(model, getAssertionUri(), resource, type, system, confidence);
+    }
+
+    /**
+     * Makes and returns a valid AIF entity object of the specified type.
+     */
+    Resource makeEntity(Resource type) {
```
craigwarsaw: Because these are supposed to be Examples as well as Tests, I would avoid making any method names in TestUtils have the same name as methods in AIFUtils.  It just makes this harder to read.  This is compounded by the fact that we use AIFUtils methods without the "AIFUtils." prefix.  So, I'd prefer method names like "makeValidBaseAIFEntity" and "makeValidNistEntity".  Likewise for makeEvent, etc.

src/test/java/com/ncc/aif/TestUtils.java
```diff
@@ -0,0 +1,466 @@
...
+        Resource relationCluster = makeClusterWithPrototype(model, getClusterUri(), relation, system);
+        return new Pair<>(relation, relationCluster);
+    }
+
+    /**
+     * Makes and returns a valid NIST-restricted hypothesis involving the specified resource(s).
+     *
+     * @param resources A set of entities, relations, and arguments that contribute to the hypothesis
+     */
+    Resource makeNistHypothesis(Resource... resources) {
```
craigwarsaw: This should go in the NistHypothesisTestUtils class.  Is there some reason why it is here instead?

src/test/java/com/ncc/aif/TestUtils.java
```diff
@@ -0,0 +1,466 @@
...
+        return hypothesis;
+    }
+
+    /**
+     * Makes and returns a valid NIST-restricted hypothesis involving the specified resource(s) and importance.
+     *
+     * @param importance the importance with which to make the hypothesis
+     * @param resources  A set of entities, relations, and arguments that contribute to the hypothesis
+     */
+    Resource makeNistHypothesis(double importance, Resource... resources) {
```
craigwarsaw: This should go in the NistHypothesisTestUtils class.  Is there some reason why it is here instead?

src/test/java/com/ncc/aif/TestUtils.java
```diff
@@ -0,0 +1,466 @@
...
+    /**
+     * Makes and returns an edge relationship between the specified event or relation and an argument filler entity.
+     *
+     * @param eventOrRelation The event or relation for which to mark the specified argument role
+     * @param type            the type of the argument
+     * @param argumentFiller  the filler (object) of the argument
+     * @param importance      the importance to mark the edge
+     * @return the created event or relation argument assertion
+     */
+    Resource makeEdge(Resource eventOrRelation, Resource type, Resource argumentFiller, double importance) {
```
craigwarsaw: Based on the main method comment (compared to the other methods), it looks like this is the only method that can make invalid AIF.  Can you confirm?  All the more important to put the word "Valid" in the other method names.

src/test/java/com/ncc/aif/TestUtils.java
nextcen-dgemoets: Nope, I just created the NistHypothesisTestUtils class at the last moment and missed moving this one.

src/test/java/com/ncc/aif/TestUtils.java
nextcen-dgemoets: Nope, I just created the NistHypothesisTestUtils class at the last moment and missed moving this one.

src/test/java/com/ncc/aif/TestUtils.java
nextcen-dgemoets: I'm not following.  How is this method suggesting that it can make invalid AIF (other that GIGO)?  As far as I know, it should make valid AIF as long as the parameters are valid (e.g., not sending a mismatched type).

craigwarsaw: The method comment of makeNistEvent says "Makes and returns a valid NIST-restricted event of the specified type and its cluster."  All the other method comments similarly say they make valid this or valid that.  Not so for makeEdge.  It just says "Makes and returns an edge relationship between the specified event or relation and an argument filler entity."  I was thinking that leaving the word "valid" out of the main method comment was intentional and did some spot checking on the use of it (e.g., it's use in invalidRelationAndEventEdge()).

So, I could definitely be wrong, but that was my thinking on my comment on this line.

nextcen-dgemoets: Hmm, the power of copy and paste, I guess.  `makeEdge()` is sort of an outlier in that it doesn't return an entity, event, relation, or hypothesis.  I probably wrote its JavaDoc solo instead of copying it and modifying it from another.

How about if I update the JavaDoc to say:
> Makes and returns a valid argument assertion between the specified event or relation and an argument filler entity.

src/test/java/com/ncc/aif/TestUtils.java
nextcen-dgemoets: Hmm, I'm not sure I agree.   Initially I did it as you suggest (e.g., `makeValidEntity`, `makeValidEvent`), but I changed it to what you see here because:
- The reality is that the only true "example code" is the stuff Ryan wrote.  Pretty much everything we've written since was for the purpose of testing functionality (mostly restrictions) and generating AIF that "shows" how to do things.  If we really wanted example code for all of the restrictions, we probably wouldn't have written these utilities because it shortens code and you'd need to dig into the TestUtils classes to see what's really happening.
- I actually thought it was a feature not a bug to use the same names when I could...if you want to make an event you call...`makeEvent`.  It's not ambiguous (arguably) because you always invoke everything through your `utils` instance...there are no static methods.
- In fact, the only reason the Nist calls have the word Nist in them (e.g., `makeNistEvent`) was because originally there was no NistTestUtils subclass, and they had different return objects.  I really should call them `makeEvent`, etc.

Maybe a compromise of sorts is to call them all `makeValidThing()` in all three classes, without the word 'Nist' or "BaseAIF' in any of the method names.  I'd be curious to hear the others' thoughts on the matter, though.

nextcen-dgemoets: I renamed everything makeValidXXX().

src/test/java/com/ncc/aif/TestUtils.java
```diff
@@ -0,0 +1,495 @@
...
+     * Makes and returns a valid NIST-restricted entity of the specified type and its cluster.
+     *
+     * @param type entity type
+     * @return a key-value Pair of the entity Resource (key) and its associated cluster Resource (value)
+     */
+    Pair<Resource, Resource> makeValidNistEntity(Resource type) {
+        Resource entity = makeEntity(model, getEntityUri(), system);
+        markJustification(addType(entity, type), getTypeAssertionJustification());
+        Resource entityCluster = makeClusterWithPrototype(model, getClusterUri(), entity, system);
+        return new Pair<>(entity, entityCluster);
```
ecurley-nextcentury: I would recommend putting this and the two lines above it into a common private static method to reduce duplicate code

src/test/java/com/ncc/aif/TestUtils.java
```diff
@@ -0,0 +1,495 @@
...
+ * Features:
+ * <p><ul>
+ * <li>Contains methods for creating a valid AIF entity, event, relation, and hypothesis objects.;</li>
+ * <li>Objects, assertions, and cluster URIs are numbered in a standard, sequential, simple way;</li>
+ * <li>Is entirely ontology-agnostic;</li>
+ * <li>Contains methods for asserting test model validity and dumping models.</li>
+ * </ul></p>
+ * Call {@link #startNewTest()} before each test to ensure a clean model.
+ */
+class TestUtils {
```
ecurley-nextcentury: I'm OK with having this be one large file, but I know some people have trouble reading such files. Was there a reason you didn't want to add each class as its own file?

src/test/java/com/ncc/aif/TestUtils.java
```diff
@@ -0,0 +1,495 @@
...
+     * Makes and returns a valid NIST-restricted entity of the specified type and its cluster with the specified
+     * cluster handle.
+     *
+     * @param type          entity type
+     * @param clusterHandle cluster handle for the entity cluster
+     * @return a key-value Pair of the entity Resource (key) and its associated cluster Resource (value)
+     */
+    Pair<Resource, Resource> makeValidNistEntity(Resource type, String clusterHandle) {
+        Pair<Resource, Resource> pair = makeValidNistEntity(type);
+        pair.getValue().addProperty(AidaAnnotationOntology.HANDLE, clusterHandle);
```
ecurley-nextcentury: Is there a reason that the handles and importances are explicit rather than some made up string/number? The main concern I have with this is that the normal makeValidNistEntity(Resource) won't create even a default handle/importance (and would thus not be valid). Let me know what you think

nextcen-dgemoets:
I've pushed an update that enables you to build and test on Linux.

craigwarsaw:
> * The ExamplesAndValidationTest classes all look very similar. I can't think of a good way to remove the redundancy, but if we could, that'd make me happy.

I'd rather address @ecurley-nextcentury's second bullet in a separate ticket (separately prioritized).

src/test/java/com/ncc/aif/TestUtils.java
```diff
@@ -0,0 +1,495 @@
...
+    String getDocumentName() {
+        return "document-" + documentCount++;
+    }
+
+    /**
+     * Returns a type assertion justification.
+     */
+    Resource getTypeAssertionJustification() {
+        if (typeAssertionJustification == null) {
+            typeAssertionJustification = makeTextJustification(model, getDocumentName(),
```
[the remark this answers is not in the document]
nextcen-dgemoets: Two reasons, one bad and one...maybe good?
The first reason is that `makeValidJustification` was added late in the game and I didn't update `getTypeAssertionJustification` to use it.  :)
The second reason is that I did kind of want the `typeAssertionJustification` to stand out a bit in the generated KB, so I kept the confidence at 0.973 and used offsets of 42 and 143.  In fact, to that end, maybe I should change the docId to something specific, like "typeAssertionDoc."  That said, it's kind of an artificial reason.  For that matter, we don't even really need `getTypeAssertionJustification` since the caller can now just call `makeValidJustification`.  I'm fine with any of these approaches.  Just let me know what you'd prefer.  Shall I just have it call `makeValidJustification` behind the scenes?

src/test/java/com/ncc/aif/TestUtils.java
```diff
@@ -0,0 +1,495 @@
...
+
+    /**
+     * Call before each test.  Returns a new, empty model with standard AIF namespaces.
+     *
+     * @return a new model with which to start a test; caller may wish to add prefixes for the ontology and annotation
+     */
+    Model startNewTest() {
+        Model model = super.startNewTest();
+        // NIST tests always need type assertion justifications
+        addSourceDocumentToJustification(super.getTypeAssertionJustification(), getDocumentName());
```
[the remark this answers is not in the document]
nextcen-dgemoets: But we need to add a source document to it in Nist.  Are you saying that `getTypeAssertionJustification` should be overloaded in Nist to add the `sourceDocument` to the return object from `super.getTypeAssertionJustification`?  I just figured I'd make sure the typeAssertionJustifcation is created and it has a source document.  I may be missing something here, but if I just remove the call to `addSourceDocumentToJustification` all of the valid tests fail.

src/test/java/com/ncc/aif/TestUtils.java
```diff
@@ -0,0 +1,495 @@
...
+        return model;
+    }
+
+    Resource makeValidJustification() {
+        return makeValidJustification("sourceDocument");
```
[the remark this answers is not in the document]
nextcen-dgemoets: Maybe it could, but I also kind of wanted the source documents to "stand out" in the generated AIF.  Also, I wasn't sure about the restrictions related to source documents, so I made a way to keep the source documents the same-- using this method, `makeValidJustification()`-- and one where you could specify a source doc in case you wanted them to be different, `makeValidJustification(String)`.

That said, if both of these reasons are weak or invalid, let me know and I can change it to use `getDocumentName()`.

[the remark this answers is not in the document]
nextcen-dgemoets: Oh, also notice that this method isn't actually used at the moment...I haven't gone back and hacked up the Nist tests to use this method.

src/test/java/com/ncc/aif/TestUtils.java
nextcen-dgemoets: It's funny you should mention that.  Somewhere along the way I noticed the same thing as you (the argument also applies to `TestUtiils` and `NistHypothesisTestUtils` although there's less code redundancy there) and I thought maybe I should just create one generic API call, `makeValidObject(Enum)` or something where you'd pass an Enum value of, essentially, ENTITY, EVENT, or RELATION.  Then it would call the appropriate private method, or perhaps if short enough, just switch off the Enum and make the call right there.  WDYT-- bad idea, too late, or go for it?

But if I don't make the above change, I can certainly do as you suggest and pull those three lines into a private method in `NistTestUtils`.

src/test/java/com/ncc/aif/TestUtils.java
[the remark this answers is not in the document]
ecurley-nextcentury: Yeah, I'm not sure our previous methodology of having a bunch of static objects was the right approach. They were added because there wasn't a utility to create valid objects on the fly. Now that there IS this great utility, ideally, the NIST examples would make use of it and all the NIST outputs would be slimmed down to just the necessities. 

I know that changing the NIST examples is probably outside the scope of this ticket, but I did think that the creation of this utility would be a general way to make generating AIF easier. With that in mind, is there a benefit to all the type justifications being the same or is it just maintaining the status quo of having a single type justification?

src/test/java/com/ncc/aif/TestUtils.java
nextcen-dgemoets: You know, if we treat `ExamplesAndValidationTest.java` as a cautionary tale, then yes, we should absolutely break this up into three files!  Having it as one file is sometimes easier to deal with, but there are valid/solid reasons to break it up (again, see `ExamplesAndValidationTest.java`).

Like you, I think I'm fine either way.  Do @craigwarsaw and @sharknc have an opinion on this?

src/test/java/com/ncc/aif/TestUtils.java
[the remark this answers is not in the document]
ecurley-nextcentury: Doesn't makeValidJustification add a sourceDocument? I would imagine we'd just use that wherever we needed a justification

src/test/java/com/ncc/aif/TestUtils.java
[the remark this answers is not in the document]
ecurley-nextcentury: Haha, well, if this method isn't needed, should we get rid of it? ;)

src/test/java/com/ncc/aif/TestUtils.java
ecurley-nextcentury: I don't think it's a bad idea or too late. As you've already done the hard work of separating the APIs, I'll leave it up to you whether you'd rather consolidate everything into a single method or just make a private, re-usable method.

ecurley-nextcentury:
> I've pushed an update that enables you to build and test on Linux.

Thanks! I'm compiling and looking at the outputs now. Will update if there are any problems.

src/test/java/com/ncc/aif/TestUtils.java
nextcen-dgemoets: Yeah, I'm glad you brought this up, because it's actually a symptom of a larger problem with the test utils.  You're right-- `NistHypothesisTestUtils` should overload `makeValidNistEntity(Resource)` to add an automatic/implicit handle and likewise `makeValidNistEvent` and `makeValidNistRelation` should overload and add an importance.

The "problem" is that it's really useful for the TA3 invalid tests to be able to create these things without the handles or importances.  See `NistHypothesisExamplesAndValidationTest::EntityClusterRequiredHandle.invalidNoHandle()` and `HypothesisEventRelationClusterImportanceValue` class.  They use the `NistTestUtils` version of makeValidEntity/Event/Relation to create these things without handles and importances, then add handles and importances as necessary to show the different invalid cases.

In general, we've created these TestUtiils so you can write the `valid()` methods very easily, but without a lot of help to the numerous `invalidXXX()` methods.  My "trick" in the above paragraph lets me have my cake and eat it too.

HOWEVER, this approach is used sporadically, and you are left with the unfortunate/bad situation where a user of the `NistHypothesisTestUtils` can create an *invalid* object using a `makeValidXXX()` method call, if they don't notice that the call comes from the parent `NistTestUtils` class (that is, if they don't add the extra `handle` or `importance` parameter).

At this stage in the game, I'm not sure what's best.  What you suggest is the pure approach, and I'd go back and update those invalid tests to create their objects mostly from scratch so that we can do everything validly except the handle/importance.  Another approach could be to rename the `NistHypoTestUtils` methods to `makeValidNistTA3XXX()`, so that you can call the TA3 version by default but still have access to the `makeValidNistXXX()` call that will be valid for Nist but not for Nist-TA3.  Ugh.

src/test/java/com/ncc/aif/TestUtils.java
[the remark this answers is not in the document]
nextcen-dgemoets: Changing the NIST examples to use the TestUtils API is in scope for this ticket, but changing them to produce better/optimal examples with less "extra stuff" is what [AIDA-720](https://nextcentury.atlassian.net/browse/AIDA-720) is about.

Like I said, having all of the type justifications makes it easier to recognize them in the generated examples, but I suspect many of you are better and reading straight AIF TTLs than I am.  ;)  I think what you're proposing is removing the `typeAssertionJustification` private member from the TestUtils class (including `getTypeAssertionJustifcation()`) and just have the examples call makeValidJustification whenever they need a justification, type assertion or otherwise.  Is this correct?  I'm on board with this, if that's what you're saying.

src/test/java/com/ncc/aif/TestUtils.java
nextcen-dgemoets: Well, the hard work is updating all of the `XXXExamplesAndValidationTest` Java files to use the new API and call `makeValidObject(TestUtils.ENTITY)` and `makeValidNistObject(TestUtils.ENTITY)`.  In theory, then, we could even make the current `makeValidEntity()` (etc.) API private...

And actually, this approach doesn't work for NistHypothesisTestUtils because entities take a String handle and events and relations take a double importance (unless we cheat and have them pass their doubles as Strings).

src/test/java/com/ncc/aif/TestUtils.java
nextcen-dgemoets: Also note that `NistExamplesAndValidationTest.:JustifyTypeAssertions.invalid()` uses `TestUtil`'s `makeValidEntity/Event/Relation()` methods to create "valid" objects that don't have justified type assertions.

src/test/java/com/ncc/aif/TestUtils.java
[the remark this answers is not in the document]
nextcen-dgemoets: Well, it isn't *used*...whether or not it's *needed* (or at least *desirable*) is a separate question.  ;)

src/test/java/com/ncc/aif/TestUtils.java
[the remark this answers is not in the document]
nextcen-dgemoets: Ah, okay, so this is coupled with your previous comment about `getTypeAssertionJustification()` calling `makeValidJustification()` (which it doesn't at the moment).  Yes, if we did that, then its polymorphic call to `makeValidJustification()` would add the `sourceDocument`, and this line could be removed.

src/test/java/com/ncc/aif/TestUtils.java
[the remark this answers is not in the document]
ecurley-nextcentury: Yeah, that's what I was thinking. Let me know if it presents any problems

src/test/java/com/ncc/aif/TestUtils.java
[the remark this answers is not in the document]
ecurley-nextcentury: Yeah, that's what I was thinking. Let me know if there are unforeseen complications

src/test/java/com/ncc/aif/TestUtils.java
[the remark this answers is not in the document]
ecurley-nextcentury: Haha, fair enough. This kinda goes along with our discussion about justifications, but if you want to be able to create valid justifications with a known, static document string, then this looks good

src/test/java/com/ncc/aif/TestUtils.java
ecurley-nextcentury: This might warrant more discussion, but I was thinking we could use what we could of the super class's functionality to create the invalid objects. Let me know if you want to discuss this Monday or some other time

ecurley-nextcentury:
Events weren't triggering the restriction if they weren't clustered. The change I pushed should fix that

src/test/java/com/ncc/aif/TestUtils.java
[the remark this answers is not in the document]
nextcen-dgemoets: Well, we just decided to get rid of `getTypeAssertionJustifcation()` in the discussion above, so the line in question is *definitely* extraneous now.  Fixed.

src/test/java/com/ncc/aif/TestUtils.java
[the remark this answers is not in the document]
nextcen-dgemoets: Done.

src/test/java/com/ncc/aif/TestUtils.java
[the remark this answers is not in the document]
nextcen-dgemoets: It now just calls `getDocumentName()`.

src/test/java/com/ncc/aif/TestUtils.java
nextcen-dgemoets: Let's discuss Monday.

src/test/java/com/ncc/aif/TestUtils.java
nextcen-dgemoets: I have moved those three lines to a private helper method, `makeValidNistObject()`.

src/test/java/com/ncc/aif/TestUtils.java
nextcen-dgemoets: I have split the classes into three separate files.

src/test/java/com/ncc/aif/TestUtils.java
nextcen-dgemoets: After discussion, we decided that a sufficient change, for now at least, is to rename the methods so that it's more explicit what you're getting, without relying on overloading methods with additional parameters.  For example, `getValidAIFEntity()` returns a valid base AIF entity, `getValidNistEntity()` returns a valid NIST-restricted entity (it is clustered and the type assertion is justified), and `getValidTA3Entity()` returns a valid NIST-restricted entity for use in TA3 (the cluster is given an importance).

Class comments were updated to reflect this.

PR #181, opened 2019-05-03
ColdStart KB now parses link assertion confidences.

### LINKED ISSUE #180: ColdStart converter does not convert confidences on KB links

rgabbard:
By default, these confidences are now required.
You can relax this requirement by setting requireLinkConfidences=False
on ColdStartKBLoader, although we currently do not expose this
in the ColdStart converter interface.

Closes #180


rgabbard:
@cedorman : Can someone from NCC merge this?  It only touches ColdStart converter code?

PR #177, opened 2019-04-29
fix platform specific character encoding build warnings

maxthomas:
current build produces the following warnings:

```
[WARNING] Using platform encoding (UTF-8 actually) to copy filtered resources, i.e. build is platform dependent!
[WARNING] File encoding has not been set, using platform encoding UTF-8, i.e. build is platform dependent!
```

fixed by this commit

PR #186, opened 2019-05-07
Aida 772

nextcen-dgemoets:
- Added `--abort` flag to validator to fail fast (after three SHACL violations), with an optional parameter that fails after [num] parameters.  When the TopBraid shacl project properly supports failing at the first violation, we can revive the [num] parameter default.  Log whether the validation was aborted, and include in summary.
- Updated README for installing 1.2 snapshot of SHACL project, and include a local maven repository used by the pom with the SHACL jar.
- Upgraded to Jena 3.11 (required by shacl 1.2 snapshot)
- Incorporated [PR-177](https://github.com/NextCenturyCorporation/AIDA-Interchange-Format/pull/177) so that it can be closed.
- Properly handle validation engine errors (exit with error code).
- Exit instead of warn if there's an unrecognized command-line option.  The warning is often lost in the output, and in most cases this means you made a mistake which might cause a long validation to occur in a way you didn't want it to.


pom.xml
```diff
@@ -78,7 +81,7 @@
         <dependency>
             <groupId>org.topbraid</groupId>
             <artifactId>shacl</artifactId>
-            <version>1.1.0</version>
+            <version>1.2.0-SNAPSHOT</version>
```
ecurley-nextcentury: We should probably just use the systemPath property: https://stackoverflow.com/a/22300875. I've tested this on my local system without performing `mvn install` on the provided .jar file and things appear to work. 
```suggestion
             <version>1.2.0-SNAPSHOT</version>
            <scope>system</scope>
            <systemPath>${project.basedir}/shacl/shacl-1.2.0-SNAPSHOT.jar</systemPath>
```

src/main/java/com/ncc/aif/ValidateAIF.java
```diff
@@ -297,7 +330,7 @@ private static boolean processArgs(String[] args, Set<ArgumentFlags> flags, Set<
 
     // Command-line argument flags
     private enum ArgumentFlags {
-        NIST, HYPO, LDC, PROGRAM, FILES, DIRECTORY, FILE_OUTPUT, PROFILING, PROGRESSIVE_PROFILING
+        NIST, HYPO, LDC, PROGRAM, FILES, DIRECTORY, FILE_OUTPUT, PROFILING, PROGRESSIVE_PROFILING, ABORT
```
ecurley-nextcentury: This is getting pretty big. It might be time to bite the proverbial bullet and make an Arguments class to house all the argument variables. In fact (it's not something I'd thought of in our discussion), but it could house the flags as well as other values like abortParam.

pom.xml
nextcen-dgemoets: As @ecurley-nextcentury and I have discussed offline, the use of `<systemPath>` is deprecated by maven.  See [this](https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#System_Dependencies) and [this](https://issues.apache.org/jira/browse/MNG-6523) for more info.  We both saw it mentioned in a [StackOverflow article](https://stackoverflow.com/questions/4955635/how-to-add-local-jar-files-to-a-maven-project) (see second answer), but I didn't use it because it was deprecated.

That said, I'm concerned that anybody using this validator will probably not see the change in the README, and things will just stop working for them.  So I'm fine with using it, but what do you think, @craigwarsaw ?

src/main/java/com/ncc/aif/ValidateAIF.java
```diff
@@ -520,6 +563,19 @@ private static ReturnCode displaySummary(int fileCount, int invalidCount, int sk
         }
     }
 
+    /**
+     * Tells the validator to "fail fast" if validation errors are detected.  Validation will terminate after
+     * [abortThreshold] validation errors are detected.  Use zero to disable failing fast.
```
ecurley-nextcentury: Is `[]` the way I should  be referencing method parameters within JavaDoc? I've been using `<code></code>`

pom.xml
nextcen-dgemoets: This has elicited further discussion that will resume with the entire team (or at least those that work with AIF).  In short, the thought is to have a branch within AIF which serves as the "AIF Developer Edition" that uses the TopBraid 1.2 snapshot, contains fail-fast and the progress monitor (AIF-791).  Then we'd keep AIF master using 1.1, with no abort flag or progress monitor.

src/main/java/com/ncc/aif/ValidateAIF.java
nextcen-dgemoets: Heh, no...`<code></code>` is correct.  The brackets were used by Ryan in AIFUtils and actually do something in Kotlin ([source](https://kotlinlang.org/docs/reference/kotlin-doc.html)).  This should probably be changed in AIFUtils also.

ecurley-nextcentury: Dang...I was kinda hoping there was a more terse form.

pom.xml
nextcen-dgemoets: Based on discussions, once we tag AIF Version 1.0, we'll merge this PR into master, along with the deprecated use of `<systemPath>`.

src/main/java/com/ncc/aif/ValidateAIF.java
nextcen-dgemoets: Based on discussions, we'll defer this and other refactoring until after we have at least preliminary results from the progress monitor and have made a little headway on TA2 processing.  The approach will be to do the full refactor-- to break out the validator command line tool from the validator class that does the actual validation.

pom.xml
craigwarsaw: Actually, I don't think we should use a tag for 1.0.  The master branch should always have the latest version of code.  Release 1.0 of AIF would be a separate long-lived branch (until we decide to no longer support it) using TopBraid 1.1.  We would have to maintain each release branch.  Fixes on the AIF 1.0 branch would have to be cherry-picked to their own feature branch off of master.  Check out this article https://docs.microsoft.com/en-us/azure/devops/repos/git/git-branching-guidance?view=azure-devops

Documenting this in AIDA Collaboration and making the AIF 1.0 branch would be part of AIDA-801.

src/main/java/com/ncc/aif/ValidateAIF.java
nextcen-dgemoets: Replaced all occurrences of `[variable]` with `<code>variable</code>`

pom.xml
nextcen-dgemoets: It turns out that `<systemPath>` doesn't actually install the jar in the `target` directory, so the validator doesn't execute.  Instead, we now create a local maven repository at `./repo` and refer to in `pom.xml`.  A Readme file was added to to `./repo`.

nextcen-dgemoets:
> I would love to see it say that it was aborted after abortThreshold failures if that is what happened rather than completely validating the model.

> it would be nice if it showed the number of KBs that were aborted early.

Unfortunately, I don't see a way to do this without modifying TopBraid.  The `MaximumNumberViolations` exception is ignored by the validation engine (the comment literally says, "ignore").  We could "guess" that the user used the abort flag if the number of violations matches the abort threshold, but there's no way to know for sure.

nextcen-dgemoets:
Based on outside conversation, it was considered acceptable that if your file has the exact number of violations as the threshold, it will still be logged and counted as an aborted validation.  So this logging/reporting has been added (and added to the README, with the above caveat).

src/main/java/com/ncc/aif/ValidateAIF.java
```diff
@@ -246,6 +252,15 @@ private static boolean processArgs(String[] args, Set<ArgumentFlags> flags, Set<
                 case "--p2": // NOTE: this flag is not documented in the README nor the Usage info
                     flags.add(ArgumentFlags.PROGRESSIVE_PROFILING);
                     break;
+                case "--abort":
+                    flags.add(ArgumentFlags.ABORT);
+                    if (i + 1 < args.length && !args[i + 1].startsWith("-")) {
```
ecurley-nextcentury: Not strictly necessary, but as this code is used in a couple places, you could use a lambda to reduce code duplication

src/main/java/com/ncc/aif/ValidateAIF.java
```diff
@@ -287,6 +303,18 @@ private static boolean processArgs(String[] args, Set<ArgumentFlags> flags, Set<
             logger.error("Please specify either file(s) or a directory of files to validate.");
             return false;
         }
+        if (abortStr != null) {
```
ecurley-nextcentury: Why not add this code above (line 258)? Readability?

src/main/java/com/ncc/aif/ValidateAIF.java
```diff
@@ -287,6 +303,18 @@ private static boolean processArgs(String[] args, Set<ArgumentFlags> flags, Set<
...
+            try {
+                // TODO: separate command-line validator from validator class
+                abortParam = Integer.parseUnsignedInt(abortStr);
+            } catch (NumberFormatException nfe) {
+                abortParam = -1;
+            }
+            if (abortParam < 1) {
```
ecurley-nextcentury: Should we also prevent 1 from being allowed? I know you have something in the README, but as it stands, 1 isn't really valid. What do you think?

src/main/java/com/ncc/aif/ValidateAIF.java
nextcen-dgemoets: Yeah, basically.  I try not to have too much code in each case of a `switch`.  So I'm just collecting whatever additional info I need immediately (such as parameters after certain options like the `[num]` after `--abort` and the files/directories after `-f` and `-d`, fail on some quick/obvious errors, and do the rest after the `switch` is over.

src/main/java/com/ncc/aif/ValidateAIF.java
nextcen-dgemoets: Hmm, I can probably argue both sides.  But it was probably re-introduced when we (or maybe just I) thought that failfast was only going to be in a "developer edition" of AIF, so we could be a little more lax about these kinds of things, since basically only AIF developers (i.e., NCC) would be using it, almost as a sandbox.  But if master is always our "best version," then I agree that num=1 should probably not be allowed.

I'll change to code to reject it on the command line, but perhaps still allow it via programmatic access (e.g., `setAbortThreshold()`) and remove the NOTE from the README.

src/main/java/com/ncc/aif/ValidateAIF.java
nextcen-dgemoets: Done.  Make sure it's what you had in mind.

src/main/java/com/ncc/aif/ValidateAIF.java
nextcen-dgemoets: By the way, I've also seen an abort threshold of 2 erroneously return a valid KB, so I've set the minimum (and default) to 3.

PR #188, opened 2019-05-08
Aida 720

craigwarsaw:
Make it easier to dump models and validation reports for unit tests.  
Check out the new README.md.  

I limited the scope of this ticket and put some of the other enhancement ideas in a separate ticket.
I am very open to better ways to do this. I thought changing code to produce the dumps is good enough for now, as long as it is properly documented.


src/test/java/com/ncc/aif/README.md
```diff
@@ -0,0 +1,41 @@
...
+## Running the tests
+
+The example unit tests are run when executing the mvn `test` target at the root directory of AIF.
+Note, the mvn `install` target also runs the `test` target.
+
+## Produce output models and reports
+
+By default, these examples do not output their models or validation reports (flags are set to false).
+You can modify these source files slightly to control which examples output their models/reports, and if so, how they output them.
+Each of the **test groups** can be separately controlled.  For example, you can output the `NistExamplesAndValidationTest.java` group models while keeping the other test groups output suppressed.
```
nextcen-dgemoets: other test groups' output

src/test/java/com/ncc/aif/README.md
```diff
@@ -0,0 +1,41 @@
...
+    - Model is dumped if the result is unexpected.
+    - Validation report is dumped if the result is invalid and unexpected.
+- **true**
+    - Model is always dumped.
+    - Validation report is dumped if the result is invalid
+
+### DUMP_TO_FILE flag
+
+- **false** (default) - If model (and validation report) is being dumped, it will go to `stdout`
+- **true** - If model (and validation report) is being dumped, it is dumped into the `target\test-dump-output` directory.  The files are named after the class and unit test method name.  For example, the `entityMissingType()` test in the `InvalidExamples` inner class of the `ExamplesAndValidationTest` class is dumped to:
```
nextcen-dgemoets: Suggest using forward slashes throughout (here and down a few lines).

src/test/java/com/ncc/aif/TestUtils.java
```diff
@@ -6,13 +6,20 @@
...
+import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.Set;
 
 import static com.ncc.aif.AIFUtils.*;
 import static org.junit.jupiter.api.Assertions.fail;
 
+import ch.qos.logback.classic.Logger;
```
nextcen-dgemoets: FWIW, we've been using IntelliJ's code formatter for consistent formatting.  One thing it does is organize imports, and it puts this import at the top (it's basically alphabetical).  If you want, I could either push an update with the code formatter changes, send you files with them and you decide what you want, or just tell you what it's changed.  It's not a log, but there are some changes it tries to make.

src/test/java/com/ncc/aif/TestUtils.java
```diff
@@ -25,11 +32,14 @@
...
 class TestUtils {
+    protected Logger logger;
 
     // Constructor parameters
     private final ValidateAIF validator;
     private final String annotationNamespace;
-    private final boolean forceDump;
+    private final boolean dumpAlways;
+    private final boolean dumpToFile;
+    private final String dumpDirectory = "test-dump-output";
```
nextcen-dgemoets: Maybe add a line between dumpToFile and dumpDirectory, as the latter isn't a constructor parameter.  For that matter, it could be a local var since it's only used in one place, but I understand why you're calling it out here.  You could also make it static to make it a full-blown class constant (and probably change it to DUMP_DIRECTORY in that case).

src/test/java/com/ncc/aif/TestUtils.java
```diff
@@ -25,11 +32,14 @@
  * Call {@link #startNewTest()} before each test to ensure a clean model.
  */
 class TestUtils {
+    protected Logger logger;
```
nextcen-dgemoets: You added the logger here, which I like.  But I think you should set the log level to INFO in the constructor here and remove that code from `declutterLogging()` as we talked about, and rename it to `initTestUtils()` or something like that.

This wouldn't prevent one of the test files to create their own logger instance (either the root or at the class level) with its own logLevel, even though right now none of the tests does that.

The point is that the TestUtils "automatically" declutter the logging a bit, but users of the utils can still override that behavior or log their own stuff independently.

src/test/java/com/ncc/aif/TestUtils.java
```diff
@@ -267,34 +280,132 @@ void testValid(String testDescription) {
         assertAndDump(testDescription, true);
     }
 
+    /**
+     * Return calling method name from test class.
+     * Looks at the calling stack for the calling test.
+     * Converts class/method from something like:
+     *  com.ncc.aif.ExamplesAndValidationTest$ValidExamples / createHierarchicalCluster
```
nextcen-dgemoets: Maybe use \<code> tags here?

craigwarsaw:
> Looks great. I generated all of the examples and they were exactly where I expected and named as I expected (that is, unambiguously). I verified on Windows and am verifying on Linux as we speak-- I'll update my review if there are any issues.
> 
> You moved the "Update tests, as necessary, to create minimal examples but still function as valid/invalid tests" requirement from AIDA-720...doesn't that mean that the examples will still have extra stuff that shouldn't be part of the example?

Probably.  There are separate tickets to address specific input from NIST.

> 
> Also, TestUtils has an unused import (`java.io.OutputStream`).

Fixed.

src/test/java/com/ncc/aif/README.md
craigwarsaw: Fixed

src/test/java/com/ncc/aif/README.md
craigwarsaw: Fixed

src/test/java/com/ncc/aif/TestUtils.java
craigwarsaw: I use Emacs and IntelliJ interchangeably.  I fixed all code using IntelliJ reformat code and optimize imports.

src/test/java/com/ncc/aif/TestUtils.java
craigwarsaw: Fixed. Changed it to class constant DUMP_DIRECTORY.

src/test/java/com/ncc/aif/TestUtils.java
craigwarsaw: I tried your requested change but now I remember why I deferred this.  Moving the setting of the log level to TestUtils results in a bunch of Jena related DEBUG output from the logger that gets dumped before the log level is set.  I think that's because the call to ValidateAIF.createForLDCOntology method as the second parameter of the call to initialize the TestUtils generates it.

src/test/java/com/ncc/aif/TestUtils.java
craigwarsaw: Fixed

src/test/java/com/ncc/aif/TestUtils.java
craigwarsaw: The ValidateAIF main gets around this by setting the loglevel to INFO, but that isn't used by the unit tests.  The split of ExamplesAndValidationTest caused the line to set the level to be duplicated, and setting the log level in the TestUtil ctor is not enough.  The real solution is to use the logback configuration file(s) to set the log level properly for the unit tests.  That can be done in a separate ticket.

craigwarsaw:
@nextcen-dgemoets - this is ready for your re-review/approval

src/test/java/com/ncc/aif/TestUtils.java
nextcen-dgemoets: Darn, you're right...it's even possible I found that during 718.  Agreed with your assessment.

src/test/java/com/ncc/aif/TestUtils.java
```diff
@@ -401,11 +400,11 @@ private void assertAndDump(String testDescription, boolean expected) {
                 dumpReport(testDescription, report);
```
nextcen-dgemoets: `testDescription` is not used in `dumpReport()` and can be removed.  Also, the JavaDoc for `dumpReport()` should  say `report` instead of `model`.  Sorry that I missed this on my first pass.

craigwarsaw: Fixed

PR #189, opened 2019-05-10
AIDA-793: Add informativeJustification and ldcTime to InterchangeOntology

ecurley-nextcentury:
Fix copy+paste error
Add classes and properties
Add test to help avoid this in the future

### DIFF
--- a/src/main/resources/com/ncc/aif/ontologies/InterchangeOntology
+++ b/src/main/resources/com/ncc/aif/ontologies/InterchangeOntology
@@ -122,6 +122,14 @@
         :KeyFrameVideoJustification
         :ShotVideoJustification ) .
 
+:LDCTime a owl:Class ;
+    rdfs:label "LDC Time" ;
+    rdfs:comment "AIF time representation for LDC format" .
+
+:LDCTimeComponent a owl:Class ;
+    rdfs:label "LDC Time Component" ;
+    rdfs:comment "Components that make up a single LDC Time" .
+
 #################################################################
 #        Properties
 #################################################################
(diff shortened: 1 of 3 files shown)


src/main/resources/com/ncc/aif/ontologies/InterchangeOntology
```diff
@@ -404,3 +412,67 @@
   rdfs:domain :PrivateData ;
   rdfs:range xsd:string ;
   rdfs:subPropertyOf owl:topDataProperty .
+
+:informativeJustification rdf:type owl:ObjectProperty ;
+  rdfs:label "informative justification" ;
+  rdfs:comment "Definitive justification for an entity or entity cluster" ;
+  rdfs:domain [
+        rdf:type owl:Class ;
+        owl:unionOf (
```
craigwarsaw: According to https://nextcentury.atlassian.net/wiki/spaces/AIDAC/pages/513736747/M18+Evaluation+Information+Including+AIF+Restrictions+and+Hypotheses and the eval plan 4.3#11, not only are entities and clusters allowed to have informative justifications, but ALSO events and relations can have them also.  They needed to be added to the unionOf, right?

ecurley-nextcentury:
Thanks, guys. Good catches! 

Apparently, an OntModel will create resources for you if they are referenced by other resources in the model. The normal Model object doesn't do this, so I've used that instead.

I have also added Event and Relation to the domain for informativeJustification

src/main/resources/com/ncc/aif/ontologies/InterchangeOntology
ecurley-nextcentury: Fixed

PR #200, opened 2019-05-24
AIDA-795: Performance enhancement

ecurley-nextcentury:
Use Jena default Model objects instead of OntModels.
Add multi-threaded validation engine based on TopBraid ValidationEngine.
Create ValidateAIFCli as an application wrapping the ValidateAIF object.

If you would like to test the performance of the threaded validator vs. the normal one, here are some .ttl files that @nextcen-dgemoets  provided:
[eddie-ttls.zip](https://github.com/NextCenturyCorporation/AIDA-Interchange-Format/files/3214732/eddie-ttls.zip)

Sorry to spam so many people as reviewers. I expect @nextcen-dgemoets will be the main reviewer, but I wanted to get enough coverage as @craigwarsaw will not likely be able to take a look before 5/29


nextcen-dgemoets:
The TTLs in the zip file have 2018 prefixes, and some of them are Seedling-based.  For the best / most accurate tests, you should change the prefixes to 2019 in all files and invoke the validator with `--ont src/main/resources/com/ncc/aif/ontologies/SeedlingOntology` instead of `--ldc` for all of the KBs except for `eric-modified.ttl` and `scalingdata-1000x300.ttl`.

src/main/java/com/ncc/aif/ValidateAIFCli.java
```diff
@@ -0,0 +1,597 @@
...
+        int skipCount = 0;
+        int abortCount = 0;
+        int fileNum = 0;
+        final StatsCollector stats = useProgressiveProfiling ?
+                new ProgressiveStatsCollector(LONG_QUERY_THRESH) : new StatsCollector(LONG_QUERY_THRESH);
+        for (File fileToValidate : filesToValidate) {
+            Date date = Calendar.getInstance().getTime();
+            logger.info("-> Validating " + fileToValidate + " at " + format.format(date) +
+                    " (" + ++fileNum + " of " + filesToValidate.size() + ").");
+            final Model dataToBeValidated = ModelFactory.createOntologyModel();
```
nextcen-dgemoets: Why not `ModelFactory.createDefaultModel()` ?

src/main/java/com/ncc/aif/ValidateAIFCli.java
```diff
@@ -0,0 +1,597 @@
...
+        logger.info("-> Validating with domain ontology(ies): " + ontologyStr);
+        if (restriction == ValidateAIF.Restriction.NIST) {
+            logger.info("-> Validating against NIST SHACL.");
+        } else if (restriction == ValidateAIF.Restriction.NIST_TA3) {
+            logger.info("-> Validating against NIST Hypothesis SHACL.");
+        }
+        if (abortSet) {
+            logger.info("-> Validation will abort after " + maxValidationErrors + " SHACL violation(s).");
+            validator.setAbortThreshold(maxValidationErrors);
+        }
```
nextcen-dgemoets: Like other features, number of threads should be logged in the "Display a summary of what we're going to do" section.  Something like:

```suggestion
        }
        if (threads != MINIMUM_THREAD_COUNT) {
            logger.info("-> Validation will occur in " + threads + " threads.");
            validator.setThreadCount(threads);
        }
```

src/main/java/com/ncc/aif/ValidateAIFCli.java
```diff
@@ -0,0 +1,597 @@
...
+ * section entitled, <i>The AIF Validator</i>.
+ */
+@CommandLine.Command(name = "validateAIF",
+        sortOptions = false,
+        synopsisHeading = "%nUsage: ",
+        descriptionHeading = "%nDescription: ",
+        optionListHeading = "%nOptions:%n",
+        description = "Used to validate Turtle files with extension .ttl",
+        mixinStandardHelpOptions = true,
+        version = "1.1.0")
```
nextcen-dgemoets: Is this automatically adding the `-V' option?  Is there any way to make it use `-v` instead?  It's the only switch/option with a capital letter in it.

src/main/java/com/ncc/aif/ValidateAIFCli.java
```diff
@@ -0,0 +1,597 @@
...
+/**
+ * A command-line AIF validator.  For details, see <a href="https://github.com/NextCenturyCorporation/AIDA-Interchange-Format">the AIF README</a>
+ * section entitled, <i>The AIF Validator</i>.
+ */
+@CommandLine.Command(name = "validateAIF",
+        sortOptions = false,
+        synopsisHeading = "%nUsage: ",
+        descriptionHeading = "%nDescription: ",
+        optionListHeading = "%nOptions:%n",
+        description = "Used to validate Turtle files with extension .ttl",
```
nextcen-dgemoets: Don't we need to mention AIDA/AIF here?  I know it can validate non-AIF, but the fact that it validates against the Interchange ontology is pretty central to its existence.

src/main/java/com/ncc/aif/ValidateAIFCli.java
```diff
@@ -0,0 +1,597 @@
...
+
+    @Option(names = "--p2", description = "Enable progressive profiling", hidden = true)
+    private boolean useProgressiveProfiling;
+
+    @Option(names = "-t", description = "Specify the number of threads to use during validation", paramLabel = "num")
+    private int threads = MINIMUM_THREAD_COUNT;
+
+    //TODO: make ArgGroup when 4.0 is stable
+    @Option(names = "-d", description = "Validate all .ttl files in the specified directory", paramLabel = "DIRNAME")
+    private File directory;
```
nextcen-dgemoets: When I tried `validateAIF -f --abort --ldc`, I got, `Expected parameter for option '-f' but found '--abort'` and the usage info, which was reasonable behavior.

But when I tried `validateAIF -d --abort --ldc`, it started to try to validate, but failed with, `No files with .ttl suffix were specified.`.

I'm not sure if this is something you can fix-- and maybe switching to ArgGroups will fix this when 4.0 is stabilized.

src/main/java/com/ncc/aif/ValidateAIFCli.java
```diff
@@ -0,0 +1,597 @@
...
+    private boolean useNISTRestriction;
+
+    @Option(names = "--nist-ta3", description = "Validate against the NIST hypothesis restrictions (implies --nist)")
+    private boolean useNISTTA3Rescriction;
+
+    @Option(names = "-o", description = "Save validation report model to a file.  KB.ttl would result in KB-report.txt.")
+    private boolean outputToFile;
+
+    @Option(names = "--abort", description = "Abort validation after [num] SHACL violations (num > 2), or 3 violations if [num] is omitted.", paramLabel = "num")
+    private int maxValidationErrors = DEFAULT_MAX_VIOLATIONS;
```
nextcen-dgemoets: The `[num]` parameter for `--abort` is supposed to be optional (and defaults to 3).
But `validateAIF --ldc -f foo.ttl --abort` results in `Missing required parameter for option '--abort' (num)`.
Similarly, `validateAIF --abort --ldc -f foo.ttl` should be valid.

src/main/java/com/ncc/aif/ValidateAIFCli.java
```diff
@@ -0,0 +1,597 @@
...
+                    ERR_MISSING_FILE_FLAG);
+        }
+        if (hasFiles && hasDirectory) {
+            throw new CommandLine.ParameterException(spec.commandLine(),
+                    ERR_TOO_MANY_FILE_FLAGS);
+        }
+    }
+
+    private void checkMinimum(int value, String name, int atLeast) {
+        // Enforce mutual exclusion for domain ontologies
```
nextcen-dgemoets: Old comment.
```suggestion
        // Enforce minimum values for certain parameters
```

src/main/java/com/ncc/aif/ValidateAIFCli.java
```diff
@@ -0,0 +1,597 @@
...
+
+    /**
+     * A statistics collector for use in profiling TopBraid-based SHACL validation.  Typical usage is to call
+     * {@link #startCollection()} and {@link #endCollection()} to bound statistics collection,
+     * then call {@link #dump(String)} to dump slow query statistics to <filename>-stats.txt.
+     * <p>
+     * This statistics collector only outputs slow queries via the {@link #dump(String)} method.  If you suspect
+     * validation will not complete due to out of memory or other error conditions, consider using {@link ProgressiveStatsCollector}.
+     */
+    private static class StatsCollector {
```
nextcen-dgemoets: This seems like a good time to move the profilers to a separate Java file.  WDYT?

src/main/java/com/ncc/aif/ValidateAIFCli.java
```diff
@@ -0,0 +1,597 @@
...
+    private static final int MINIMUM_MAX_VIOLATIONS = 3;
+    // Profiling
+    private static final int LONG_QUERY_THRESH = 2000;
+    // Threading
+    private static final String THREAD_COUNT_STRING = "Thread count";
+    private static final int MINIMUM_THREAD_COUNT = 1;
+
+    //=-=-=-=-=-=-=-=-=-=-=-=-=-=-=
+    // Command Line Arguments
+    // ----------------------------
```
nextcen-dgemoets: Some formatting things in Usage output:
- Previously, option descriptions were designed to fit 80-char screens.  If it's possible, it would be nice to preserve this behavior.  There are line breaks being added which are making it look bad in 80 characters.
- One dash options (`-v`) are indented differently from two-dash switches (`--abort`).  Any idea why?

One dash options (-v) are indented differently from two-dash switches (--abort).

src/main/java/com/ncc/aif/ValidateAIFCli.java
```diff
@@ -0,0 +1,597 @@
...
+ */
+@CommandLine.Command(name = "validateAIF",
+        sortOptions = false,
+        synopsisHeading = "%nUsage: ",
+        descriptionHeading = "%nDescription: ",
+        optionListHeading = "%nOptions:%n",
+        description = "Used to validate Turtle files with extension .ttl",
+        mixinStandardHelpOptions = true,
+        version = "1.1.0")
+public class ValidateAIFCli implements Callable<Integer> {
```
nextcen-dgemoets: _Putting this here since it refers to an unmodified file._

I think the README.md file should be updated to bring the usage message in sync with program output:
- Add -V / --version and -t [num]
- Update -h / --help
- Could you add backtick in --pm?  My bug.

Also, it occurs to me that the beginning of the README only indirectly mentions the validator CLI.  That first bullet should refer to the appropriate section in the README ("The AIF Validator").

src/main/java/com/ncc/aif/ValidateAIF.java
```diff
@@ -596,8 +179,10 @@ public void setProgressMonitor(String id) {
         try {
             this.progressMonitor = new AIFProgressMonitor(filename);
         } catch (IOException e) {
-            logger.warn("Could not open progress monitor filename {}.  Writing progress to StdOut.", filename);
+            // Default constructor for AIFProgressMonitor makes use of logger. Add logging there
```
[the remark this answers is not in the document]
ecurley-nextcentury: As ValidateAIF no longer has a logger, I was attempting to indicate that the logger used by AIFProgressMonitor would be used. Can you think of another way of saying this? Should we even log anything as this is the only place we use a logger?

src/main/java/com/ncc/aif/ValidateAIF.java
```diff
@@ -614,6 +199,37 @@ public void setAbortThreshold(int abortThreshold) {
...
+    /**
+     * Return the current executor if one exists, null o/w
+     * @return the current executor if one exists, null o/w
+     */
+    public ExecutorService getExecutor() {
+        return executor;
+    }
+
+    //TODO: remove this
+    public List<Future<ThreadedValidationEngine.ValidationMetadata>> getValidationMetadata() {
```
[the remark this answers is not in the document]
ecurley-nextcentury: This can't be removed until we get rid of the performance test. Let me know what you think about removing it now. Do you think we could use the futures for the performance monitor?

src/main/java/com/ncc/aif/ValidateAIF.java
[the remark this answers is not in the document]
nextcen-dgemoets: Now that it no longer has a logger, how about just putting it to stderr?  Or we could throw the exception and have the caller decide if it wants to log and/or instantiate a `AIFProgressMonitor` with default ctor.

src/main/java/com/ncc/aif/ValidateAIFCli.java
ecurley-nextcentury: I would have, but I'm waiting for direction on package structure. Should there be a validator package?

src/main/java/com/ncc/aif/ValidateAIFCli.java
ecurley-nextcentury: I'm not sure this will be fixed in 4.0, but the behavior with -d is a bug with picocli. We can get around this bug by making the -d switch act the same as -f. This would allow multiple directories, though.

Given a conversation with @nextcen-dgemoets, we're just going to leave it as-is for now

src/main/java/com/ncc/aif/ValidateAIFCli.java
nextcen-dgemoets: Based on discussion, we're deferring this.

src/main/java/com/ncc/aif/ValidateAIF.java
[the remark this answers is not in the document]
nextcen-dgemoets: Gotcha-- just change comment to say something like remove this when `ValidatorPerformanceTest` is removed.  Although I'm not opposed to keeping the test (and therefore this method).

src/main/java/com/ncc/aif/ValidateAIFCli.java
```diff
@@ -0,0 +1,597 @@
...
+    private int maxValidationErrors = DEFAULT_MAX_VIOLATIONS;
+
+    @Option(names = "--pm", description = "Enable progress monitor that shows ongoing validation progress")
+    private boolean useProgressMonitor;
+
+    @Option(names = "-p", description = "Enable profiling", hidden = true)
+    private boolean useProfiling;
+
+    @Option(names = "--p2", description = "Enable progressive profiling", hidden = true)
+    private boolean useProgressiveProfiling;
```
nextcen-dgemoets: I tried profiling a multi-threaded validation.  It didn't go well due to memory usage:
```
HAL t=3, 26GB, profiler (-p), ran OOM after about 6 minutes.  Memory usage doubled from 13G to 26GB in about 70s.
HAL t=3, 26GB, profiler (-p2), ran OOM after about 9 minutes.  Similar memory spike (10-26G in about 60s).
```

It's not necessarily a requirement that they co-exist, but it's something potentially to investigate and if nothing else, to document.

src/main/java/com/ncc/aif/ValidateAIF.java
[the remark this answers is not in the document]
ecurley-nextcentury: Fixed

src/main/java/com/ncc/aif/ValidateAIFCli.java
ecurley-nextcentury: Fixed

src/main/java/com/ncc/aif/ValidateAIFCli.java
ecurley-nextcentury: Fixed

src/main/java/com/ncc/aif/ValidateAIFCli.java
ecurley-nextcentury: Fixed

src/main/java/com/ncc/aif/ValidateAIFCli.java
ecurley-nextcentury: Fixed

src/test/java/com/ncc/aif/ValidateCLITest.java
```diff
@@ -0,0 +1,129 @@
...
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+public class ValidateCLITest {
+    private static final boolean SHOW_OUTPUT = false;
+    PrintStream oldOut;
+    PrintStream oldErr;
+    ByteArrayOutputStream baos;
+
```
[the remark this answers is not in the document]
ecurley-nextcentury: Fixed

src/test/java/com/ncc/aif/ValidateCLITest.java
```diff
@@ -0,0 +1,129 @@
...
+    }
+
+    @BeforeEach
+    void createCLI() {
+        baos = new ByteArrayOutputStream();
+        PrintStream out = new PrintStream(baos);
+        System.setOut(out);
+        System.setErr(out);
+    }
+
```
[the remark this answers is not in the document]
ecurley-nextcentury: Added these 3 tests

README.md
```diff
@@ -66,12 +66,15 @@ Usage:  <br>
...
 |`--abort [num]` | Abort validation after `[num]` SHACL violations (num > 2), or three violations if `[num]` is omitted. |
-|`--pm | Enable progress monitor that shows ongoing validation progress |
-|`-f FILE ...` | validate the specified file(s) with a `.ttl` suffix |
+|`--pm` | Enable progress monitor that shows ongoing validation progress |
+|`-o` | Save validation report model to a file.  `KB.ttl` would result in `KB-report.txt`. Output defaults to stderr. |
+|`-t num` | Specify the number of threads to use during validation |
 |`-d DIRNAME` | validate all `.ttl` files in the specified directory |
+|`-f FILE ...` | validate the specified file(s) with a `.ttl` suffix |
+|`-h, --help` | This help and usage text |
+|`-h, --help` | This help and usage text |
```
nextcen-dgemoets: Extra `-h` line, and could you add the equals sign to `-f`, `-d`, and `-t` options?

src/main/java/com/ncc/aif/ValidateAIFCli.java
nextcen-dgemoets: Could you fix the SNAPSHOT number in the README?
And in the first bullet:

> This ontology can be validated using the SHACL constraints file in      `src/main/resources/com/ncc/aif/aida_ontology.shacl`.

how about adding something like, " by the AIF Validator."

src/main/java/com/ncc/aif/ValidateAIF.java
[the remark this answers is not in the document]
nextcen-dgemoets: Looks like you addressed this.  Resolving.

src/main/java/com/ncc/aif/ValidateAIFCli.java
```diff
@@ -0,0 +1,597 @@
...
+    private boolean useLDCOntology;
+
+    @Option(names = "--program", description = "Validate against the program ontology")
+    private boolean useProgramOntology;
+
+    @Option(names = "--ont", description = "Validate against the OWL-formatted ontolog(ies) at the specified filename(s)",
+            paramLabel = "FILE", arity = "1..*")
+    private List<File> customOntologies;
+
+    //TODO: make ArgGroup when 4.0 is stable
```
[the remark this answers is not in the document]
nextcen-dgemoets: Looks like this is fixed.  Resolving.

src/main/java/com/ncc/aif/ValidateAIFCli.java
nextcen-dgemoets: Looks like this is fixed.  Resolving.

src/main/java/com/ncc/aif/ValidateAIFCli.java
nextcen-dgemoets: Looks like this is fixed.  Resolving.

src/main/java/com/ncc/aif/ValidateAIFCli.java
nextcen-dgemoets: Looks like this is fixed.  Resolving.

src/main/java/com/ncc/aif/ValidateAIF.java
```diff
@@ -614,6 +199,37 @@ public void setAbortThreshold(int abortThreshold) {
...
+     *
+     * @param threadCount number of threads to use during validation
+     */
+    public void setThreadCount(int threadCount) {
+        if (threadCount < 1) {
+            throw new IllegalArgumentException("Number of threads must be greater than or equal to 1.");
+        }
+        if (threadCount > 1 && (executor == null || executor.getPoolSize() != threadCount)) {
+            executor = new ThreadPoolExecutor(threadCount, threadCount,0L, TimeUnit.MILLISECONDS,
+                    new LinkedBlockingQueue<Runnable>());
```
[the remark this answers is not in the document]
nextcen-dgemoets: Looks like this is addressed.  Resolving.

src/main/java/com/ncc/aif/ValidateAIFCli.java
ecurley-nextcentury: We're still on 1.1.0, right? That's what's in the POM.

src/main/java/com/ncc/aif/ThreadedValidationEngine.java
```diff
@@ -0,0 +1,206 @@
...
+import java.net.URI;
+import java.util.*;
+import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Future;
+import java.util.function.Predicate;
+import java.util.stream.Collectors;
+
+public class ThreadedValidationEngine extends ValidationEngine {
```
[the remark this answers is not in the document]
ecurley-nextcentury: Let's talk offline to figure out what you'd like here and what's sufficient for this PR

src/main/java/com/ncc/aif/ValidateAIFCli.java
nextcen-dgemoets: Oh for goodness sakes, every now and then I mix the AIF and TopBraid versions.  Yes, the README is correct.

I agree that they don't have to use the validator to use the SHACL file, but I disagree that the current wording is sufficient, in that I believe the validator CLI should be mentioned in the intro.  When those bullets were written, there might not even have been a validator class written yet (or if there was, it was rudimentary).  Now it's a full-featured command-line tool (with programmatic access) for validating AIF, which we recommend users of AIF use to validate their KBs.  It's a core "resource to support [AIF]," which is what the opening paragraph and bullets are describing.  I'm open to other ways of mentioning it, perhaps in a separate bullet?

nextcen-dgemoets:
FYI, I ran a directory of 202 KBs through the AIDA-795 validator and it took 9m31s on my laptop to validate the entire directory (single-threaded).  They were all valid.
I ran them with t=4 and it took 5min33s, but the penultimate file was deemed invalid ("Unsupported justification type").  I ran it again with t=4, and they were all valid the second time.

Also of note perhaps, none of their files use any prefixes.

src/main/java/com/ncc/aif/ValidateAIFCli.java
ecurley-nextcentury: Added bullet to README concerning what this repository contains. Let me know if it's sufficient

src/main/java/com/ncc/aif/ThreadedValidationEngine.java
```diff
@@ -0,0 +1,206 @@
...
+        return result;
+    }
+
+    @Override
+    public void setFocusNodeFilter(Predicate<RDFNode> value) {
+        super.setFocusNodeFilter(value);
+        focusNodeFilter = value;
+    }
+
+    private void throwMaxiumNumberViolationsIfReached(int violations) {
```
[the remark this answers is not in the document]
ecurley-nextcentury: fixed

src/main/java/com/ncc/aif/ThreadedValidationEngine.java
```diff
@@ -0,0 +1,206 @@
...
+        }
+    }
+
+    @Override
+    public Resource createResult(Resource type, Constraint constraint, RDFNode focusNode) {
+        Resource result = threadModel.get().createResource(type);
+        result.addProperty(SH.resultSeverity, constraint.getShapeResource().getSeverity());
+        result.addProperty(SH.sourceConstraintComponent, constraint.getComponent());
+        result.addProperty(SH.sourceShape, constraint.getShapeResource());
+        if(focusNode != null) {
```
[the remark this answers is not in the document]
ecurley-nextcentury: fixed

src/main/java/com/ncc/aif/ThreadedValidationEngine.java
```diff
@@ -0,0 +1,206 @@
...
+        for(Shape shape : rootShapes) {
+            validationMetadata.add(executor.submit(getTask(shape, i++, executor)));
+        }
+
+        Map<String, Model> models = new HashMap<>();
+        int violations = 0;
+        for (Future<ValidationMetadata> future : validationMetadata) {
+            ValidationMetadata md = future.get();
+            models.computeIfAbsent(md.threadName, key -> md.model);
+            throwMaxiumNumberViolationsIfReached(violations += md.violations);
```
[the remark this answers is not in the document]
ecurley-nextcentury: fixed

src/main/java/com/ncc/aif/ValidateAIFCli.java
```diff
@@ -0,0 +1,597 @@
...
+    private boolean useNISTTA3Rescriction;
+
+    @Option(names = "-o", description = "Save validation report model to a file.  KB.ttl would result in KB-report.txt.")
+    private boolean outputToFile;
+
+    @Option(names = "--abort", description = "Abort validation after [num] SHACL violations (num > 2), or 3 violations if [num] is omitted.", paramLabel = "num")
+    private int maxValidationErrors = DEFAULT_MAX_VIOLATIONS;
+
+    @Option(names = "--pm", description = "Enable progress monitor that shows ongoing validation progress")
+    private boolean useProgressMonitor;
```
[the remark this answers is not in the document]
ecurley-nextcentury: fixed

README.md
ecurley-nextcentury: fixed

src/main/java/com/ncc/aif/ValidateAIFCli.java
nextcen-dgemoets: Looks good-- thanks.  Resolving.

README.md
```diff
@@ -63,16 +65,15 @@ Usage:  <br>
 | ----------- | ----------- |
 |`--ldc`    | validate against the LDC ontology |
 |`--program`| validate against the program ontology |
-|`--ont FILE ...` | validate against the OWL-formatted ontolog(ies) at the specified filename(s) |
+|`--ont=FILE ...` | validate against the OWL-formatted ontolog(ies) at the specified filename(s) |
 |`--nist` | validate against the NIST restrictions |
 |`--nist-ta3` | validate against the NIST hypothesis restrictions (implies `--nist`) |
-|`--abort [num]` | Abort validation after `[num]` SHACL violations (num > 2), or three violations if `[num]` is omitted. |
+|`--abort=[num]` | Abort validation after `[num]` SHACL violations (num > 2), or three violations if `[num]` is omitted. |
```
nextcen-dgemoets: It's `--abort[=num]` like in the Usage.

src/main/java/com/ncc/aif/ValidateAIFCli.java
```diff
@@ -126,7 +126,8 @@ public Integer convert(String value) {
     @Option(names = "-o", description = "Save validation report model to a file.  KB.ttl would result in KB-report.txt.")
     private boolean outputToFile;
 
-    @Option(names = "-t", description = "Specify the number of threads to use during validation", paramLabel = "num")
+    @Option(names = "-t", description = "Specify the number of threads to use during validation. As the threaded validator" +
+            " doesn't currently support adding a progress monitor, this disables the --pm option.", paramLabel = "num")
```
nextcen-dgemoets: Nice.  I think this should probably go in the README also, but I don't object to leaving it out as we intend to add multi-threaded progress monitoring "soon."

README.md
ecurley-nextcentury: Copying and pasting is fun!

src/main/java/com/ncc/aif/ValidateAIFCli.java
ecurley-nextcentury: I would argue that it IS in the README. Copy+paste!

README.md
ecurley-nextcentury: fixed

nextcen-dgemoets: Oh my, how did I miss that?  I was looking for it in `--pm` in the README for some reason.  My bad.

src/main/java/com/ncc/aif/ValidateAIFCli.java
nextcen-dgemoets: Oh my, how did I miss that?  I was looking for it in `--pm` in the README for some reason.  My bad.

src/main/java/com/ncc/aif/ThreadedValidationEngine.java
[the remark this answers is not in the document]
nextcen-dgemoets: So, it's not crashing, but it's also not aborting after the correct number of violations.  I seem to get, pretty reliably, twice the number of violations in my report than `--abort` specifies, regardless of my value of t (>1).

As a side effect, the final report isn't reporting that validation was aborted.

I don't know that it's *absolutely critical* that validation abort after *exactly* `[num]` violations, as long as it's documented and the final summary reflects that it aborted.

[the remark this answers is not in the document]
ecurley-nextcentury: Yeah, we don't have absolute control over validation errors as we're multi-threaded. I'll add something to the cli to report an abort.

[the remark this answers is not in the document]
nextcen-dgemoets: Would it be silly to *remove* violations from the report model (essentially truncate up to the error threshold)?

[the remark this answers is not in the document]
ecurley-nextcentury: Haha, it seems to go against the spirit of the feature. Basically, you just want to fail once you've found the number of violations. If we've already found the violations, it seems odd to not report them.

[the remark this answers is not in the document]
nextcen-dgemoets: I agree, but it saves having to document the (perhaps unexpected) behavior or do any special-case handling in the CLI.  I guess I don't feel strongly either way...so far.  ;-)

src/main/java/com/ncc/aif/ThreadedValidationEngine.java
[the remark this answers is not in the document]
nextcen-dgemoets: To kickstart that discussion, here's a starting point from which we can add/subtract:
- Basics...what the class does.
- Differences between it and TopBraid's base class
  - Usage, performance, limitations (e.g., progress monitor), issues (e.g., `setValidationErrorBatch`)
- Very high level implementation?
- Does it work in cases where restrictions are all in the data model (e.g., `shapesModel = null`)?
- As for `ValidationMetadata` : quick description of what it is / how it's used.  Basic.

src/main/java/com/ncc/aif/ThreadedValidationEngine.java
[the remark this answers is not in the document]
ecurley-nextcentury: fixed

src/main/java/com/ncc/aif/ThreadedValidationEngine.java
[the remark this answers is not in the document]
ecurley-nextcentury: fixed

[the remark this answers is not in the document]
nextcen-dgemoets: I think we might want to add a little more as part of AIDA-816, but this is certainly fine for this ticket.

src/main/resources/com/ncc/aif/version.properties
```diff
@@ -0,0 +1 @@
+version=${project.version}
```
nextcen-dgemoets: It looks like these properties files are missing a newline.

ecurley-nextcentury: Good catch. Fixed

nextcen-dgemoets:
FYI, after seeing the TA2 file validate fairly quickly this morning on my laptop (10m, or 7m with t=4), I re-ran the tests on HAL 2.0.  I verified my command history to ensure I was running the same commands as before (ldc/nist/output/threads).  The only additional step I performed-- other than getting the latest from 795-- was to wipe out my Maven cache.  I got different results today:
```
HAL t=0, 26GB, took 8.5 mins (mem topped at 8 GB)
HAL t=3, 26GB, took 5 mins (mem topped at 9.6 GB)
HAL t=4, 26GB, took 4.5 mins 
HAL t=8, 26GB, took 4 mins (mem topped at 9.4 GB)
```
I don't really have a full explanation, but I like these numbers a lot better.  I think we all agree that more testing is warranted on the MT validator, but I feel a lot better about it now than I did a few days ago.

By the way, I never did see CPU usage get particularly high, even when running with 8 threads.  It's not like all 8 cores were working hard validating their particular rules.  It may be that a few rules in particular were causing the bottlenecks

src/main/java/com/ncc/aif/ThreadedValidationEngine.java
```diff
@@ -0,0 +1,245 @@
...
+//        if (monitor != null) {
+//            monitor.beginTask("Validating " + rootShapes.size() + " shapes", rootShapes.size());
+//        }
+
+        int i = 0;
+        for (Shape shape : rootShapes) {
+            validationMetadata.add(executor.submit(getTask(shape, i++, executor)));
+        }
+
+        // Go through all futures and get md for those that have completed
```
sharknc: change `md` in comment to `validation metadata`

src/main/java/com/ncc/aif/ValidateAIFCli.java
```diff
@@ -0,0 +1,649 @@
...
+
+        final ReturnCode returnCode = displaySummary(fileNum + nonTTLcount, invalidCount, skipCount + nonTTLcount, abortCount);
+        return returnCode.ordinal();
+    }
+
+    //TODO: make ArgGroup when 4.0 is stable
+    private void checkOntMutex() {
+        // Enforce mutual exclusion for domain ontologies
+        boolean hasCustom = customOntologies != null;
+        if (!(useProgramOntology || useLDCOntology || hasCustom)) {
```
sharknc: Does picocli not support this out of the box through configurations?

src/test/java/com/ncc/aif/ValidatorPerformanceTest.java
```diff
@@ -0,0 +1,107 @@
...
+
+//TODO: remove me
+public class ValidatorPerformanceTest {
+    private static final String ROOT = "../VERDI/MockTA1Algo/";
+    private static final String SAMPLE = ROOT + "R103.Sample.2019.05.14/";
+    private static final String KB = SAMPLE + "R103.kb.ttl";
+    private static final String SMALL_TA1 = SAMPLE + "IC00120RO.parent.ttl";
+    private static final String MEDIUM_TA1 = SAMPLE + "IC0011WX8.parent.ttl";
+    private static final String BIG_TA1 = ROOT + "R103.2019.05.23/IC0011TIJ.parent.ttl";
+    private static final String GAIA_KB = "/home/HQ/ecurley/Documents/AIDA/2019.05.17-AIDA-795-Threading/big-gaia-sansprivate.ttl";
```
sharknc: You may want to either update the path of this or put a comment above stating that this test requires the `big-gaia-sansprivate.ttl` with a location on where to find it.

src/main/java/com/ncc/aif/ThreadedValidationEngine.java
```diff
@@ -0,0 +1,245 @@
...
+        }
+
+        return report;
+    }
+
+    private Callable<ValidationMetadata> getTask(Shape shape, int id, ExecutorService executor) {
+        return () -> {
+            long start = System.currentTimeMillis();
+            boolean nested = SHACLScriptEngineManager.begin();
+//            if (monitor != null) {
```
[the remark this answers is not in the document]
ecurley-nextcentury: Yeah, this is for the progress monitor if/when we implement a multi-threaded version

src/main/java/com/ncc/aif/ThreadedValidationEngine.java
```diff
@@ -0,0 +1,245 @@
...
+                            validateNodesAgainstConstraint(filtered, constraint);
+                        } catch (MaximumNumberViolations e) {
+                            if (!executor.isShutdown()) {
+                                executor.shutdownNow();
+                            }
+                        }
+                    }
+                }
+            }
+//            if (monitor != null) {
```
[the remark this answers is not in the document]
ecurley-nextcentury: Yeah, this is for the progress monitor if/when we implement a multi-threaded version

src/main/java/com/ncc/aif/ThreadedValidationEngine.java
```diff
@@ -0,0 +1,245 @@
...
+     *
+     * @param executor {@link ExecutorService} to send jobs to
+     * @return an instance of sh:ValidationReport in the results Model
+     * @throws InterruptedException when {@link Future#get()} experiences {@link InterruptedException}
+     * @throws ExecutionException when {@link Future#get()} experiences {@link ExecutionException}
+     */
+    public Resource validateAll(ExecutorService executor) throws InterruptedException, ExecutionException {
+        List<Shape> rootShapes = shapesGraph.getRootShapes();
+        //TODO: Add monitor support for threaded validator. Experience NPE with AIFProgressMonitor
+//        if (monitor != null) {
```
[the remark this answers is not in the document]
ecurley-nextcentury: Yeah, this is for the progress monitor if/when we implement a multi-threaded version

src/main/java/com/ncc/aif/ValidateAIFCli.java
ecurley-nextcentury: 4.0 supports it, but it's currently in Alpha and has a bug where the exclusivity is caught, but the wrong error message is produced. Hopefully, we'll change all these to ArgGroups when 4.0 is stable

src/test/java/com/ncc/aif/ValidatorPerformanceTest.java
ecurley-nextcentury: Good idea. Will comply

src/main/java/com/ncc/aif/ThreadedValidationEngine.java
ecurley-nextcentury: fixed

src/test/java/com/ncc/aif/ValidatorPerformanceTest.java
ecurley-nextcentury: fixed

PR #205, opened 2019-05-30
AIDA-827: Allow clusters to be in hypothesis:

ecurley-nextcentury:
Use default Jena models instead of OntModel.
Add SameAsCluster to subgraphContains.
Add KEsInModelMustBeReferencedByHypothesis.validWithClusterAndMembership test.

### DIFF
--- a/src/test/java/com/ncc/aif/NistTA3ExamplesAndValidationTest.java
+++ b/src/test/java/com/ncc/aif/NistTA3ExamplesAndValidationTest.java
@@ -280,6 +280,8 @@ public class NistTA3ExamplesAndValidationTest {
             }
         }
 
+        @Disabled("Disabled: This shape is not implemented correctly. A KE should consist of cluster, all membership nodes," +
+                " all member nodes, all type assertions for member nodes, and all justifications for type assertions for member nodes")
         @Nested
         class KEsInModelMustBeReferencedByHypothesis {
             Resource relation;
--- a/src/test/java/com/ncc/aif/NistTA3ExamplesAndValidationTest.java
+++ b/src/test/java/com/ncc/aif/NistTA3ExamplesAndValidationTest.java
@@ -5,6 +5,7 @@ import ch.qos.logback.classic.Logger;
 import com.google.common.collect.ImmutableSet;
 import org.apache.commons.lang3.tuple.ImmutablePair;
 import org.apache.jena.rdf.model.Model;
+import org.apache.jena.rdf.model.ResIterator;
 import org.apache.jena.rdf.model.Resource;
 import org.junit.jupiter.api.*;
 import org.junit.jupiter.api.TestInstance.Lifecycle;
@@ -283,6 +284,7 @@ public class NistTA3ExamplesAndValidationTest {
         class KEsInModelMustBeReferencedByHypothesis {
             Resource relation;
             Resource relationEdge;
+            Resource relationCluster;
 
             @BeforeEach
             void setup() {
@@ -290,6 +292,7 @@ public class NistTA3ExamplesAndValidationTest {
                         LDCOntology.GeneralAffiliation_ArtifactPoliticalOrganizationReligiousAffiliation,
                         103.0);
                 relation = relationPair.getKey();
+                relationCluster = relationPair.getValue();
                 relationEdge = utils.makeValidTA3Edge(relation,
                         LDCOntology.GeneralAffiliation_ArtifactPoliticalOrganizationReligiousAffiliation_EntityOrFiller,
                         entity, 102.0);
@@ -303,9 +306,17 @@ public class NistTA3ExamplesAndValidationTest {
             }
 
             @Test
-            void valid() {
+            void validWithoutCluster() {
                 utils.makeValidTA3Hypothesis(entity, relation, relationEdge, event, eventEdge);
-                utils.testValid("NISTHypothesis.valid: All KEs in model must be referenced by hypothesis");
+                utils.testValid("NISTHypothesis.validWithoutClusters: All KEs in model must be referenced by hypothesis");
+            }
+
+            @Test
+            void validWithClusterAndMembership() {
+                ResIterator it = model.listSubjectsWithProperty(AidaAnnotationOntology.CLUSTER_PROPERTY, relationCluster);
+                Assertions.assertTrue(it.hasNext(), "Unable to find expected cluster membership");
+                utils.makeValidTA3Hypothesis(entity, relation, relationEdge, event, eventEdge, relationCluster, it.nextResource());
+                utils.testValid("NISTHypothesis.validWithClusterAndMembership: All KEs in model must be referenced by hypothesis");
             }
         }
 
--- a/src/main/java/com/ncc/aif/ValidateAIF.java
+++ b/src/main/java/com/ncc/aif/ValidateAIF.java
@@ -6,6 +6,7 @@ import com.google.common.base.Charsets;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.io.CharSource;
 import com.google.common.io.Resources;
+import org.apache.jena.ontology.OntModel;
 import org.apache.jena.rdf.model.*;
 import org.apache.jena.riot.RDFDataMgr;
 import org.apache.jena.riot.RDFFormat;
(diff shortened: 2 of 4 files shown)


ecurley-nextcentury:
> Everything looks fine, but...why is the change from OntModels to default models in this PR? It doesn't seem to belong here, since it's in [AIDA-795](https://nextcentury.atlassian.net/browse/AIDA-795), which is close to merging into master, and will be further verified (and then applied to v1.0.x) in [AIDA-810](https://nextcentury.atlassian.net/browse/AIDA-810).

No reason to not include it and it made my testing go faster :D

nextcen-dgemoets:
I get the part about the validation speed, I just don't think it should be merged into master out of turn.  I would have been in support of a separate task just about the model switch (which would have been approved/merged by now).  But lacking that, I think it should wait until 795, which is where the default model's benefits were discovered, as is a performance-related task.

nextcen-dgemoets:
In other words, the reason not to include it is that it's a fairly important change and has nothing to do with this task.

ecurley-nextcentury:
I agree it's an important change, but I'm less confident that 795 is going to get in soon. I would also point out that the change to default model has nothing to do with 795. I'll remove it from this PR, though.

nextcen-dgemoets:
If 795 isn't going to get in soon, then maybe we should talk to Craig about a possible new ticket to make the switch to default model.  Maybe we can even rewrite/repurpose AIDA-810.

src/test/java/com/ncc/aif/NistTA3ExamplesAndValidationTest.java
```diff
@@ -303,9 +306,17 @@ void invalid() {
...
                 utils.makeValidTA3Hypothesis(entity, relation, relationEdge, event, eventEdge);
-                utils.testValid("NISTHypothesis.valid: All KEs in model must be referenced by hypothesis");
+                utils.testValid("NISTHypothesis.validWithoutClusters: All KEs in model must be referenced by hypothesis");
+            }
+
+            @Test
+            void validWithClusterAndMembership() {
+                ResIterator it = model.listSubjectsWithProperty(AidaAnnotationOntology.CLUSTER_PROPERTY, relationCluster);
+                Assertions.assertTrue(it.hasNext(), "Unable to find expected cluster membership");
+                utils.makeValidTA3Hypothesis(entity, relation, relationEdge, event, eventEdge, relationCluster, it.nextResource());
```
craigwarsaw: This generated AIF that I wouldn't have expected.  It put a blank node in the hypothesis subgraph (attached).  I think the ClusterMemberships should be named.

[KEsInModelMustBeReferencedByHypothesis.validWithClusterAndMembership.ttl.txt](https://github.com/NextCenturyCorporation/AIDA-Interchange-Format/files/3239038/KEsInModelMustBeReferencedByHypothesis.validWithClusterAndMembership.ttl.txt)

ecurley-nextcentury: ClusterMembership nodes have always been blank. Are you suggesting that this ticket be used to make a change like naming ClusterMemberships?

ecurley-nextcentury:
> Either we might want to relax this constraint OR we should change the message to be explicit about what we are not allowing to be unreferenced in the hypothesis (rather than using the confusing term "KE"). For that matter, the name of this unit test is probably confusing as well.

What would you suggest? I'm inclined to just remove the constraint as NIST doesn't need it.

craigwarsaw:
> > Everything looks fine, but...why is the change from OntModels to default models in this PR? It doesn't seem to belong here, since it's in [AIDA-795](https://nextcentury.atlassian.net/browse/AIDA-795), which is close to merging into master, and will be further verified (and then applied to v1.0.x) in [AIDA-810](https://nextcentury.atlassian.net/browse/AIDA-810).
> 
> No reason to not include it and it made my testing go faster :D

Actually, there is a reason not to include the change from OntModel to default model in this branch.  Assuming we can cherry pick certain commits that are intentionally targeted for a 1.0.x patch release, we want the changes to be limited to the specific task so as to minimize potential regression errors.

IMO, if this wasn't intended for a patch release, these kind of changes are subject to a judgment call and, in general, well-justified improvements can get in OR be pushed to a separate ticket and prioritized separately.

ecurley-nextcentury:
> > > Everything looks fine, but...why is the change from OntModels to default models in this PR? It doesn't seem to belong here, since it's in [AIDA-795](https://nextcentury.atlassian.net/browse/AIDA-795), which is close to merging into master, and will be further verified (and then applied to v1.0.x) in [AIDA-810](https://nextcentury.atlassian.net/browse/AIDA-810).
> > 
> > 
> > No reason to not include it and it made my testing go faster :D
> 
> Actually, there is a reason not to include the change from OntModel to default model in this branch. Assuming we can cherry pick certain commits that are intentionally targeted for a 1.0.x patch release, we want the changes to be limited to the specific task so as to minimize potential regression errors.
> 
> IMO, if this wasn't intended for a patch release, these kind of changes are subject to a judgment call and, in general, well-justified improvements can get in OR be pushed to a separate ticket and prioritized separately.

Agreed. As both of these changes are slated for a patch release, we probably should have pushed them together. Especially as the ont change in 795 isn't the only thing in its commit, so we are unable to cherry pick.

craigwarsaw:
> > > > Everything looks fine, but...why is the change from OntModels to default models in this PR? It doesn't seem to belong here, since it's in [AIDA-795](https://nextcentury.atlassian.net/browse/AIDA-795), which is close to merging into master, and will be further verified (and then applied to v1.0.x) in [AIDA-810](https://nextcentury.atlassian.net/browse/AIDA-810).
> > > 
> > > 
> > > No reason to not include it and it made my testing go faster :D
> > 
> > 
> > Actually, there is a reason not to include the change from OntModel to default model in this branch. Assuming we can cherry pick certain commits that are intentionally targeted for a 1.0.x patch release, we want the changes to be limited to the specific task so as to minimize potential regression errors.
> > IMO, if this wasn't intended for a patch release, these kind of changes are subject to a judgment call and, in general, well-justified improvements can get in OR be pushed to a separate ticket and prioritized separately.
> 
> Agreed. As both of these changes are slated for a patch release, we probably should have pushed them together. Especially as the ont change in 795 isn't the only thing in its commit, so we are unable to cherry pick.

Sorry if I wasn't clear.  I don't think they should have pushed together.  I think there is more testing we need to do before we can release the ont model to default model change and I'd like to get 1.0.1 out earlier than that.  The fix for allowing clusters in hypothesis is targeted for 1.0.1.  I just created a new candidate release for 1.0.2 which includes the ont model to default model change and the enhanced unit tests.

craigwarsaw:
> > Either we might want to relax this constraint OR we should change the message to be explicit about what we are not allowing to be unreferenced in the hypothesis (rather than using the confusing term "KE"). For that matter, the name of this unit test is probably confusing as well.
> 
> What would you suggest? I'm inclined to just remove the constraint as NIST doesn't need it.

I agree that NIST doesn't need it.  So we can remove it rather than unnecessarily "fixing" it or trying to make a clearer for an "incomplete" constraint.  

To be clear, the constraint NIST needs is to ensure that the hypothesis subgraph can contain clusters and clustermemberships.  NIST has said that we do not need to invalidate a hypothesis file that has extra KEs in the file that are not in the subgraph.  Right?

nextcen-dgemoets:
> To be clear, the constraint NIST needs is to ensure that the hypothesis subgraph can contain clusters and clustermemberships. NIST has said that we do not need to invalidate a hypothesis file that has extra KEs in the file that are not in the subgraph. Right?

I was kind of wondering how exactly that was left, too.  They definitely said they didn't absolutely need the restriction (their queries will still work), but I also got the impression that they consider anything in the file as part of their evaluation, so maybe it behooves performers not to include extras stuff in there that could throw off results.  If nothing else, I don't think NIST said, "yeah, go ahead and remove that restriction."  It would be nice to remove it, though, especially if it's causing us problems or more work.

craigwarsaw:
> > To be clear, the constraint NIST needs is to ensure that the hypothesis subgraph can contain clusters and clustermemberships. NIST has said that we do not need to invalidate a hypothesis file that has extra KEs in the file that are not in the subgraph. Right?
> 
> I was kind of wondering how exactly that was left, too. They definitely said they didn't absolutely need the restriction (their queries will still work), but I also got the impression that they consider anything in the file as part of their evaluation, so maybe it behooves performers not to include extras stuff in there that could throw off results. If nothing else, I don't think NIST said, "yeah, go ahead and remove that restriction." It would be nice to remove it, though, especially if it's causing us problems or more work.

The evaluation plan is clear that they shouldn't have extras in there.  NIST basically said that

1. We didn't have to invalidate files that included extras.
2. We didn't have to put in the extra work to remove the existing restriction.

Getting into the implementation, it seems these are contradictory.  The problem is the validation message is confusing, so we either have to clarify the validation message or remove that restriction. Since NIST gave us the choice, erring on the side of being lenient allows people's submissions to be queried for evaluation rather than possibly erroneously or confusingly being reported as invalid.  If removing the restriction is easy, let's do that.

src/test/java/com/ncc/aif/NistTA3ExamplesAndValidationTest.java
ecurley-nextcentury: Per discussion, have disabled these tests until KE restriction is implemented correclty

PR #220, opened 2019-06-05
AIDA-853

nextcen-dgemoets:
Bumped version number to 1.0.1.


nextcen-dgemoets:
Those two files already say 1.0.1.  That was done on the front end, because we don't bother setting those files to 1.0.1-SNAPSHOT.

nextcen-dgemoets:
That said, maybe moving forward it's better not to update those two files until we update the POM to 1.0.x?  I don't feel strongly either way.

PR #228, opened 2019-06-13
Add optional uri_ref argument to mark_as_possible_cluster_member()

### LINKED ISSUE #227: Add optional uri_ref to mark_as_possible_cluster_member()
This is extending what was done in #88 to this new function.

For BBN, we find it useful to diff our output when running regression tests, so avoiding blank nodes is important as they are printed in a non-deterministic order.

rogerbock:
Closes #227


rogerbock:
This is a pretty minor change that only will affect behavior if you take advantage of a new optional parameter.  Any chance this could get accepted?

PR #233, opened 2019-06-18
Adding m18 submission checker script

nmj101:
Linked to ticket AIDA-871

### Description:
This is designed to be a standalone python3 script that can be sent out to performers so that they can run it on their archives and check if their archive is submit ready. Submit ready means that the archive is structured properly for the AIF M18 batch validation process. 

#### To test: 

- Ensure that the task type was determined correctly; 
- that the output is clear as to which task type the archive was checked for; 
- that the program does not raise false positives in where it states an archive is submit ready when it should not be. 

#### To run:
`python3 aida_m18_submission_checker.py <path/to/archive>`

### DIFF
--- a/scripts/aida_m18_submission_checker.py
+++ b/scripts/aida_m18_submission_checker.py
@@ -1,0 +1,370 @@
+'''
+This script is designed to parse an archive [.zip, .tgz, .tar.gz] and determine if it is properly structured, and will
+work with Next Century's validator.
+
+To run:
+python3 aida_m18_submission_checker.py <path/to/archive>
+
+Output will be printed to stdout and/or stderr
+'''
+import sys                  #for sys.args
+from sys import exit as exit
+import os                   #for extention access (basename)
+import tarfile, zipfile     #for archive access
+import re                   #for regex
+
+
+'''
+Globals
+'''
+
+#variable to store if we use tar or zip method to access the file
+#-1 = error. cannot continue
+#0 = .tar.gz, or .tgz
+#1 = .zip
+FILETYPE = -1
+
+#task types to validate for
+TASK_TYPES = ["1a", "1b", "2", "3"]
+
+#list of names of items in the archive - to be populated after determining what filetype the archive is
+ARCHIVE_NAMES = []
+
+
+def eprint(*args, **kwargs):
+    """simple function to print to stderr
+
+    :param args: data to print
+    :param kwargs: keywords to pass in print args
+    """
+    #*****************
+    # REQUIRES PYTHON 3+
+    #*****************
+    print(*args, file=sys.stderr, **kwargs)
+
+
+def do_TA1a_check():
+    """
+    Check the archive
+    TASK 1a Directory structure:
+    <TA1aperformer>_<run>
+        NIST
+            <document_id>.ttl                  (1 to X)
+		INTER-TA
+		    <document_id>.ttl                  (0 to X)
+
+    Ignores non ttl files.
+    Prints number of ttl files found, and number of ttl files that are in need to be checked.
+
+    :return: True if it appears submit ready, False otherwise
+    :rtype: bool
+    """
+
+    ttls_found_count = 0
+    ttls_error_warning_count = 0
+    ttls_valid_count = 0
+
+    for name in ARCHIVE_NAMES:
+
+        #split into levels
+        path_items = re.split('/', name)
+
+#TODO CHECK TA1A AND TA1B FOR IF PATH_ITEMS[-1] IS EVER NOT THE LOCATION OF THE .TTL FILE NAME
+
+        if '.ttl' in path_items[-1]:
+
+            ttls_found_count = ttls_found_count + 1
+
+            #validate that the ttl file is in an acceptable directory
+            #rule: ttl needs to be in 2nd level -- aka 3rd item in list['name', 'nist', 'valid.ttl']
+            #rule: ttl needs to be in directory named nist, or inter-ta
+            try:
+                if ((path_items[-2].upper() == 'NIST' or path_items[-2].upper() == 'INTER-TA') and len(path_items) == 3):
+                    ttls_valid_count = ttls_valid_count + 1
+                else:
+                    ttls_error_warning_count = ttls_error_warning_count + 1
+            except:
+                ttls_error_warning_count = ttls_error_warning_count + 1
+
+    print("\nTA_1a Archive Validator report:")
+    print("INFO: TTL files found: " + str(ttls_found_count))
+    print("INFO: TTL files in valid locations: " + str(ttls_valid_count))
+    if ttls_error_warning_count > 0:
+        print("ERROR: TTL files found in an invalid location: " + str(ttls_error_warning_count))
+        print(".ttl files must be located in either the <run_ID>/NIST/ or <run_id>/INTER-TA directory")
+
+    if ttls_found_count > 0 and ttls_error_warning_count == 0:
+        return True
+    return False
+
+
+def do_TA1b_check():
+    """
+    Task 1b directory structure:
+    <TA1performer>_<run>
+        NIST
+            hypothesisID			(1 to Y)
+                <document_id>.ttl 		(1 to X)
+
+    Ignores non ttl files.
+    Prints number of ttl files found, and number of ttl files that are in need to be checked.
+
+    :return: True if it appears submit ready, False otherwise
+    :rtype: bool
+    """
+    ttls_found_count = 0
+    ttls_error_warning_count = 0
+    ttls_valid_count = 0
+
+    for name in ARCHIVE_NAMES:
+
+        #split into levels
+        path_items = re.split('/', name)
+
+        if '.ttl' in path_items[-1]:
+
+            ttls_found_count = ttls_found_count + 1
+
+            #validate that the ttl file is in an acceptable directory
+            #rule: ttl needs to be in 3rd level -- aka 4th item in list['name', 'nist', 'hypothesis', 'valid.ttl']
+            #rule: ttl needs to be in a subdirectory within the nist directory
+            try:
+                if path_items[-3].upper() == 'NIST' or len(path_items) == 4:
+                    ttls_valid_count = ttls_valid_count + 1
+                else:
+                   ttls_error_warning_count = ttls_error_warning_count + 1
+            except:
+                ttls_error_warning_count = ttls_error_warning_count + 1
+
+    print("\nTA_1b Archive Validator report:")
+    print("INFO: TTL files found: " + str(ttls_found_count))
+    print("INFO: TTL files in valid locations: " + str(ttls_valid_count))
+    if ttls_error_warning_count > 0:
+        print("ERROR: TTL files found in an invalid location: " + str(ttls_error_warning_count))
+        print("ttl files must be located the appropriate hypothesis_id subdirectory of the <run_ID>/NIST/ directory. i.e. <run_id>/NIST/<hypothesis_id>/file.ttl")
+
+    if ttls_found_count > 0 and ttls_error_warning_count == 0:
+        return True
+    return False
+
+
+def do_TA2_check():
+    """
+    Task 2 directory structure:
+    <TA1performer>_<run>(-<TA1performer>_<run>).<TA2performer>_<run>
+        NIST
+            <TA1performer>_<run>(-<TA1performer>_<run>).<TA2performer>_<run>.ttl	(1)
+        INTER-TA
+            <TA1performer>_<run>(-<TA1performer>_<run>).<TA2performer>_<run>.ttl	(0 or 1)
+
+
+    Ignores non ttl files.
+    Prints number of ttl files found, and number of ttl files that are in need to be checked.
+
+    we do not check for all NIST standards set by the document, only for basic directory structure and a NIST/INTER-TA directory
+    :return: True if it appears submit ready, False otherwise
+    :rtype: bool
+    """
+
+    ttls_found_count = 0
+    ttls_error_warning_count = 0
+    ttls_valid_count = 0
+
+    for name in ARCHIVE_NAMES:
+
+        #split into levels
+        path_items = re.split('/', name)
+
+        if '.ttl' in path_items[-1]:
+
+            ttls_found_count = ttls_found_count + 1
+
+            #validate that the ttl file is in an acceptable directory
+            #rule: ttl needs to be in 2nd level -- aka 3rd item in list['name', 'nist', 'valid.ttl']
+            #rule: ttl needs to be in directory named nist, or inter-ta
+            try:
+                if ((path_items[-2].upper() == 'NIST' or path_items[-2].upper() == 'INTER-TA') and len(path_items) == 3):
+                    ttls_valid_count = ttls_valid_count + 1
+                else:
+                    ttls_error_warning_count = ttls_error_warning_count + 1
+            except:
+                ttls_error_warning_count = ttls_error_warning_count + 1
+
+
+    print("\nTA_2 Archive Validator report:")
+    print("INFO: TTL files found: " + str(ttls_found_count))
+    print("INFO: TTL files in valid locations: " + str(ttls_valid_count))
+    if ttls_error_warning_count > 0:
+        print("ERROR: TTL files found in an invalid location: " + str(ttls_error_warning_count))
+        print(".ttl files must be located in either the <run_ID>/NIST/ or <run_id>/INTER-TA directory")
+
+    if ttls_found_count > 0 and ttls_error_warning_count == 0:
+        return True
+    return False
+
+
+def do_TA3_check():
+    """
+    Task 3 directory structure:
+    <TA1performer>_<run>(-<TA1performer>_<run>).<TA2performer>_<run>(-<TA2performer>_<run>).<TA3performer>_<run>
+        <TA1performer>_<run>(-<TA1performer>_<run>).<TA2performer>_<run>(-<TA2performer>_<run>).<TA3performer>_<run>.<SIN ID>.<SIN frame ID>.<H followed by three digits>.ttl			(1 to X)
+
+    Ignores non ttl files.
+    Prints number of ttl files found, and number of ttl files that are in need to be checked.
+
+    we do not check for all NIST standards set by the document, only for basic directory structure.
+    :return: True if it appears submit ready, False otherwise
+    :rtype: bool
+    """
+
+    ttls_found_count = 0
+    ttls_error_warning_count = 0
+
+    for name in ARCHIVE_NAMES:
+        print(name)
+        #split into levels
+        path_items = re.split('/', name)
+        print(path_items)
+        if '.ttl' in path_items[-1]:
+
+            ttls_found_count = ttls_found_count + 1
+
+            #validate that the ttl file is in an acceptable directory
+            #rule: ttl needs to be in 2nd level -- aka 3rd item in list['name', 'nist', 'valid.ttl']
+            #rule: ttl needs to be in directory named nist, or inter-ta
+
+            if len(path_items) is not 2:
+                ttls_error_warning_count = ttls_error_warning_count + 1
+
+    print("\nTA_3 Archive Validator report:")
+    print("INFO: TTL files found: " + str(ttls_found_count))
+    print("INFO: TTL files in valid locations: " + str(ttls_found_count - ttls_error_warning_count))
+    if ttls_error_warning_count > 0:
+        print("ERROR: TTL files found in an invalid location: " + str(ttls_error_warning_count))
+        print(".ttl files must be located in either the <run_ID>/NIST/ or <run_id>/INTER-TA directory")
+
+    if ttls_found_count > 0 and ttls_error_warning_count == 0:
+        return True
+    return False
+
+
+def get_task_type(file, args):
+    """
+
+    :param file: the archive file
+    :param args: the runtime arguments for the program -- used to determine the task type
+    :return: int corresponding to the task type (see TASK_TYPE)
+    :rtype: int
+    """
+
+    #user did not tell us what task type it was, we must determine it ourselves.
+    #do this with counting the number of .'s in the filename (get rid of .tar.gz first though)
+    basename = os.path.basename(file)
+
+    #get rid of extra count if extention is .tar.gz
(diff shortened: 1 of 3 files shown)


scripts/aida_m18_submission_checker.py
```diff
@@ -0,0 +1,370 @@
...
+    global ARCHIVE_NAMES
+
+    #ensure the archive type is a valid filetype and set the global FILETYPE value
+    if not validate_archive_filetype(sys.argv[1]) and FILETYPE >= 0:
+        eprint("FATAL ERROR: Archive filetype unknown. Please use .zip, .tgz, or .tar.gz")
+        exit(False)
+
+    #get names of items in the archive
+    if '-v' in sys.argv: print ("Gathering archive items...")
+    archive = zipfile.ZipFile(sys.argv[1], 'r', allowZip64=True) if FILETYPE else tarfile.open(sys.argv[1], mode='r')
```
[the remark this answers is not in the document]
nmj101: I did the filetype thing in case there was an additional archive library that was needed to read the archive. However, because python does not support if (int):  as I intended, that would work just as well.

scripts/aida_m18_submission_checker.py
```diff
@@ -0,0 +1,370 @@
...
+                ttls_error_warning_count = ttls_error_warning_count + 1
+
+    print("\nTA_1a Archive Validator report:")
+    print("INFO: TTL files found: " + str(ttls_found_count))
+    print("INFO: TTL files in valid locations: " + str(ttls_valid_count))
+    if ttls_error_warning_count > 0:
+        print("ERROR: TTL files found in an invalid location: " + str(ttls_error_warning_count))
+        print(".ttl files must be located in either the <run_ID>/NIST/ or <run_id>/INTER-TA directory")
+
+    if ttls_found_count > 0 and ttls_error_warning_count == 0:
```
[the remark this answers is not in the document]
nmj101: Is there ever a time that passing extra .ttl files in an incorrect location is necessary? I feel like ensuring they don't have any ttls in the wrong places is the best move.

scripts/aida_m18_submission_checker.py
```diff
@@ -0,0 +1,370 @@
...
+    :return: int corresponding to the task type (see TASK_TYPE)
+    :rtype: int
+    """
+
+    #user did not tell us what task type it was, we must determine it ourselves.
+    #do this with counting the number of .'s in the filename (get rid of .tar.gz first though)
+    basename = os.path.basename(file)
+
+    #get rid of extra count if extention is .tar.gz
+    dot_count = basename.count('.') if not '.tar.gz' in basename else basename.count('.') - 1
```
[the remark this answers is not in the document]
nmj101: changed to basename.endswith('.tar.gz') rather than in

scripts/aida_m18_submission_checker.py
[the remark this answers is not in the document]
nmj101: made it a global and changed get_task_type to use the global rather than take the filename as an argument

nmj101:
changes for commit: https://github.com/NextCenturyCorporation/AIDA-Interchange-Format/pull/233/commits/baef01da22e877c1cc0e020577dc33e1f480a4d3 


+ added a ARCHIVE_FILE_NAME global to store filename >> solves sys.argv[1]
+ changed from print to logging
- removed eprint << MAKES IT COMPATABLE WITH PYTHON2.7 too?
- removed import comments
- removed check for -v
+ added runtime input/argument validation
+ fixed logic with validate_archive_filetype() usage in main()
+ fixed the if (int) statements << rookie mistake, didnt know python 
+ fixed the a = a + 1 << ditto
+ made the directory checks for 'NIST' and 'INTER-TA' case sensitive
+ fixed all " '.ext' in name " for appropriate name.endswith('.ext')
+ added a valid_ttls_count in do_TA3_check() for consistency
+ fixed 1b logic 'and' instead of 'or'
~ contemplated duplicates - no changes made
- removed random debugging statements
+ added try catch in main for opening and reading file >> in case user renames a .txt to .zip, or archive is empty file
- changed get_task_type to use the global ARCHIVE_FILE_NAME and not an argument
- validate_archive_filetype() removed argument and rewritten to use the global filename
-   changed from: file_ext = os.path.splitext(file)[-1].lower()
+   to: if ARCHIVE_FILE_NAME.endswith('.tar.gz')... 
+ added a README.md -- this will probably need changes

scripts/m18_submission_checker/aida_m18_submission_checker.py
```diff
@@ -0,0 +1,383 @@
...
+        return False
+
+    #get names of items in the archive
+    try:
+        logging.info("Checking archive... this may take a moment for large files.")
+        archive = zipfile.ZipFile(ARCHIVE_FILE_NAME, 'r', allowZip64=True) if FILETYPE else tarfile.open(ARCHIVE_FILE_NAME, mode='r')
+        ARCHIVE_NAMES = archive.namelist() if FILETYPE == 1 else archive.getnames()
+    except:
+        #expected errors: (zipfile.BadZipFile, tarfile.ReadError), but catch all of them anyways
+        logging.critical("Error thrown attempting to read/open the archive. Please use 'zip' or 'tar' to create your archive.")
```
sharknc: I would be more explicit in the text of this statement letting the user know that the only valid submissions are .tar.gz, .tgz, and .zip. This will really never be thrown once you fix the small bug with the if statement above.

scripts/m18_submission_checker/aida_m18_submission_checker.py
```diff
@@ -0,0 +1,383 @@
...
+        logging.error("Missing archive file as argument.\nTo run: python3 aida_m18_submission_checker.py <path/to/archive>")
+        return False
+    if not os.path.isfile(sys.argv[1]):
+        logging.error("Argument is not a file.\nTo run: python3 aida_m18_submission_checker.py <path/to/archive>")
+        return False
+
+    ARCHIVE_FILE_NAME = str(sys.argv[1])
+
+    #ensure the archive type is a valid filetype and set the global FILETYPE value
+    if validate_archive_filetype() < 0:
```
sharknc: This is a bug, the `validate_archive_filetype()` returns a boolean, not an int. Update this to be 
`if not validate_archive_filetype():`

scripts/m18_submission_checker/aida_m18_submission_checker.py
```diff
@@ -0,0 +1,383 @@
...
+            try:
+                if ((path_items[-2] == 'NIST' or path_items[-2] == 'INTER-TA') and len(path_items) == 3):
+                    ttls_valid_count = ttls_valid_count + 1
+                else:
+                    ttls_error_warning_count += 1
+            except:
+                ttls_error_warning_count += 1
+
+    logging.info("\033[1mTA_1a\033[0m Archive Validator report:")
+    logging.info("TTL files found: " + str(ttls_found_count))
```
sharknc: I would change this log statement text slightly. TTL files found can be interpreted various ways. I would recommend changing to
`Total number of .ttl files found in submission: x`

scripts/m18_submission_checker/aida_m18_submission_checker.py
```diff
@@ -0,0 +1,383 @@
...
+    logging.info("TTL files in valid locations: " + str(ttls_valid_count))
```
sharknc: Like the comment above I would change the wording on this slightly to something like:
`Total number of .ttl files to be validated: x`

scripts/m18_submission_checker/aida_m18_submission_checker.py
```diff
@@ -0,0 +1,383 @@
...
+    if ttls_error_warning_count > 0:
+        logging.error("TTL files found in an invalid location: " + str(ttls_error_warning_count))
```
sharknc: This probably should be `logging.warning` instead of an error.

scripts/m18_submission_checker/aida_m18_submission_checker.py
```diff
@@ -0,0 +1,383 @@
...
+        logging.error(".ttl files must be located in either the <run_ID>/NIST/ or <run_id>/INTER-TA directory")
+
+    if ttls_found_count > 0 and ttls_error_warning_count == 0:
```
sharknc: I think this was an issue in my previous comment, but I am not sure you want to check that the warning count must be zero here. Just because .ttl files are not in a valid location does not mean that their submission won't be validated. It just means those files will not be included in the validation. @craigwarsaw we should probably talk about this more.

scripts/m18_submission_checker/aida_m18_submission_checker.py
```diff
@@ -0,0 +1,383 @@
...
+            #rule: ttl needs to be in directory named NIST, or INTER-TA
+            try:
+                if ((path_items[-2] == 'NIST' or path_items[-2] == 'INTER-TA') and len(path_items) == 3):
+                    ttls_valid_count = ttls_valid_count + 1
+                else:
+                    ttls_error_warning_count += 1
+            except:
+                ttls_error_warning_count += 1
+
+    logging.info("\033[1mTA_1a\033[0m Archive Validator report:")
```
sharknc: I am on the fence about the bolding in the text here. Is this universally compatible? It is a nice feature, I just don't want it to break if someone tries to run this on Windows.

scripts/m18_submission_checker/aida_m18_submission_checker.py
```diff
@@ -0,0 +1,383 @@
...
+                ttls_valid_count += 1
+            else:
+                ttls_error_warning_count += 1
+
+    logging.info("\033[1mTA_3\033[0m Archive Validator report:")
+    logging.info("TTL files found: " + str(ttls_found_count))
+    logging.info("TTL files in valid locations: " + str(ttls_valid_count))
+    if ttls_error_warning_count > 0:
+        logging.error("TTL files found in an invalid location: " + str(ttls_error_warning_count))
+        logging.error(".ttl files must be located in either the <run_ID>/NIST/ or <run_id>/INTER-TA directory")
```
sharknc: This log statement is incorrect for the TA3 submission task type.

scripts/m18_submission_checker/aida_m18_submission_checker.py
nmj101: further research indicates that the escape sequence does not play nicely with all windows versions. I'll figure a different way to make sure the user knows which TA type is being checked for.

scripts/m18_submission_checker/aida_m18_submission_checker.py
nmj101: the error thrown by this block is due to 
1. nothing in the archive
2. an invalid archive format (i.e. some text file renamed to .tgz)

we previously check for appropriate file extension and warn the user

sharknc: Okay, the only reason I suggested this was due to the failure in one of my test cases that I described below. It made it through your validation check when it should have failed and this was the messages I received.

scripts/m18_submission_checker/aida_m18_submission_checker.py
nmj101: removed bolding and changed output style

scripts/m18_submission_checker/aida_m18_submission_checker.py
nmj101: this is strange, ill test this out more.

scripts/m18_submission_checker/aida_m18_submission_checker.py
```diff
@@ -0,0 +1,295 @@
...
+    # log archive report
+    logging.info("M18 Submission Checker Report for: \n\t{0}".format(archive_file_name))
+    logging.info("Total .ttl files found in submission archive: {0}".format(archive_status['total']))
+    logging.info("Total number of .ttl files to be validated: {0}".format(archive_status['valid']))
+    #if they have ttls located in invalid locations, print the warning
+    if archive_status['invalid'] > 0:
+        logging.warning("Total number of .ttl files located in invalid locations and will not be validated: {0}".format(archive_status['invalid']))
+        logger_print_level = logging.warning if archive_status['valid'] > 0 else logging.error
+        logger_print_level(warnings_dict[task_type])
+    if archive_status['valid'] > 0:
```
craigwarsaw: The SUCCESS message should only be shown if archive_status['valid'] > 0 and archive_status['invalid'] == 0 (in other words, there are no TTLs in unexpected locations).

There should be one additional case:
if archive_status['valid'] > 0 and archive_status['invalid'] > 0:
log a warning
"PARTIAL SUCCESS! {0} is organized according to task {1} rules and can be submitted but there are .ttl files in unexpected locations that will not be validated"

scripts/m18_submission_checker/README.md
```diff
@@ -0,0 +1,22 @@
+# See section 9 of the NIST AIDA 2019 Evaluation Plan V1.3 document for proper archive naming convention, formatting, and archive file types. #
+
+This script is designed to help check if an archive is submit ready for M18 evaluation. 
+Submit ready means that the archive is structured properly for the AIF M18 batch validation process.
+
+This script checks that the ttl files are located in a valid directory structure (i.e. not in the base directory when expected to be in a /NIST/ directory).
+
```
craigwarsaw: Add
This script does not do a complete check against all of the submission archives naming rules.  It checks for the proper submission file name extension and the number of periods in the submission file name.

scripts/m18_submission_checker/README.md
```diff
@@ -0,0 +1,22 @@
+# See section 9 of the NIST AIDA 2019 Evaluation Plan V1.3 document for proper archive naming convention, formatting, and archive file types. #
+
+This script is designed to help check if an archive is submit ready for M18 evaluation. 
+Submit ready means that the archive is structured properly for the AIF M18 batch validation process.
```
craigwarsaw: "Submit ready" means that the archive is structured properly for the Next Century's automated AIF M18 batch validation process.

scripts/m18_submission_checker/aida_m18_submission_checker.py
craigwarsaw: See my comment below.  I distinguish between a SUCCESS with all files in the right location and a PARTIAL SUCCESS where it can be validated but there are some files in the wrong location.

scripts/m18_submission_checker/aida_m18_submission_checker.py
nmj101: will write out diagram of output flow in next commit

scripts/m18_submission_checker/aida_m18_submission_checker.py
```diff
@@ -0,0 +1,295 @@
...
+        # if ".ttl file in a subdirectory of /NIST/" then ttls_under_nist_subdir is not none (possible 1b)
+        ttls_under_nist_subdir = re.compile(".*\/NIST\/\S*\/\S*\.ttl", re.IGNORECASE).search(member_names_str)
+
+        if ttls_under_nist_dir is not None and ttls_under_nist_subdir is None:
+            return '1a'
+        elif ttls_under_nist_subdir is not None and ttls_under_nist_dir is None:
+            return '1b'
+        else:
+            # ttls found in /NIST/ and a subdirectory of /NIST/, so we cannot determine if 1a or 1b.
+            logging.warning("Cannot distinguish between 1a or 1b due to .ttl files existing at the same level as a subdirectory of /NIST/. \nContinuing as 1a.")
```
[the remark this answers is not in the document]
nmj101: next commit will not change the way it distinguishes between 1a and 1b, however it WILL error upon no ttl files in a NIST directory or no NIST directory.

scripts/m18_submission_checker/aida_m18_submission_checker.py
nmj101: implemented the distinction, double check proper usage on next commit

scripts/m18_submission_checker/aida_m18_submission_checker.py
```diff
@@ -0,0 +1,381 @@
...
+                required_exist = (ttl_valid_count_dict[k] > 0)
+
+        if required_exist == None:
+            required_exist = False
+
+    # log archive report
+    logging.info("Total .ttl files found in submission archive: {0}".format(ttls_total_count))
+    logging.info("Total number of .ttl files to be validated: {0}".format(count_to_be_validated))
+
+    #if invalid dont exsist, dont print this line
```
craigwarsaw: misspelled "exist"

scripts/m18_submission_checker/aida_m18_submission_checker.py
```diff
@@ -0,0 +1,381 @@
...
+    if (((task_type == '1a' or task_type == '2') and required_exist) or (count_to_be_validated > 0 and required_exist == None)) and count_not_to_be_validated == 0:
+        logging.info("SUCCESS! {0} is organized according to task {1} rules and is ready to be submitted!".format(archive_file_name, task_type))
+        return True
+    else:
+        # point user in the right direction when theres some sort of unexpected thing happening
+        logging.info("For task {0}: {1}".format(task_type, warnings_dict[task_type]))
+
+        #and also do the correct status message
+        if count_to_be_validated == 0 or not required_exist:
+            logging.error("FAILURE! {0} is not organized according to task {1} rules, or has 0 files where 1 or more is expected.".format(archive_file_name, task_type))
```
craigwarsaw: This line is erroneously reached in the following situation:
1. Task 3 submission
2. X .ttl files in top level folder (valid)
3. Y .ttl files in a subdirectory (invalid)

This should be a PARTIAL SUCCESS because the validator will be able to validate X files but not Y files

Also:
1. Task 1b submission
2. X .ttl files in NIST/hypo1 (valid)
3. Y .ttl files in NIST/hypo2 (valid)
4. Z .ttl files in INTER-TA (invalid)

This should be a PARTIAL SUCCESS because the validator will be able to validate X+Y files but not Z files
Actually 4 above, if Z files are anywhere else other than NIST/hypo folder (e.g., in top level), then it fails when it should be a partial success.

I **think** if you switch the if and the elif then it may work right, but you should double check that.

scripts/m18_submission_checker/aida_m18_submission_checker.py
```diff
@@ -0,0 +1,381 @@
...
+
+    try:
+        if ((path_items[-2] == 'NIST' or path_items[-2] == 'INTER-TA') and len(path_items) == 3):
+            return True
+    except:
+        return False
+    return False
+
+
+def do_TA1b_check(path_items):
```
sharknc: I created a submission in the following format

RUN_ID/NIST/Hypothesis_1/{2 .ttl files}
RUN_ID/NIST/Hypothesis_2/{2 .ttl files}
RUN_ID/INTER-TA/{6 .ttl files}

This fails. I would expect it to be a PARTIAL SUCCESS because the NIST folder with nested Hypothesis folders with .ttls is valid. You can get this example in s3 at `aida-validation/m18_checker/Test_1b_1a-Valid.zip1`

scripts/m18_submission_checker/aida_m18_submission_checker.py
```diff
@@ -0,0 +1,381 @@
...
+    # rule: ttl needs to be in 3rd level -- aka 4th item in list['name', 'NIST', 'hypothesis', 'valid.ttl']
+    # rule: ttl needs to be in a subdirectory within the NIST directory
+    try:
+        if path_items[-3] == 'NIST' and len(path_items) == 4:
+            return True
+    except:
+        return False
+    return False
+
+def do_TA3_check(path_items):
```
sharknc: I created a submission in the following format

RUN_ID/NIST/{2 .ttl files}
RUN_ID/INTER-TA/{6 .ttl files}
RUN_ID/{2 .ttl files}

This fails. I would expect it to be a PARTIAL SUCCESS because there are two .ttl files in the RUN_ID folder. You can get this example in s3 at `aida-validation/m18_checker/Test_3.Test_3.Test_3a-Valid.zip`

scripts/m18_submission_checker/aida_m18_submission_checker.py
sharknc: I just realized that my comments below describe these same issues. Sorry for the duplicate comments.

scripts/m18_submission_checker/aida_m18_submission_checker.py
```diff
@@ -0,0 +1,381 @@
...
+    archive_member_names = get_archive_member_names(archive_file_name, archive_file_type)
+
+    if archive_member_names is None:
+        return False
+
+    task_type = get_task_type(archive_file_name, archive_member_names)
+    if task_type == None:
+        return False
+
+    logging.info("Based on the number of periods in the file name, this archive is assumed to be a Task {0} submission.".format(task_type))
```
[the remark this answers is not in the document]
nmj101: added individual log/warning statements for each task type

scripts/m18_submission_checker/aida_m18_submission_checker.py
```diff
@@ -0,0 +1,381 @@
...
+    :returns: True on Success or Partial success, False otherwise
+    """
+
+    logging.info("M18 Submission Checker Report for: \n\t{0}".format(archive_file_name))
+
+    # get totals
+    count_to_be_validated = sum(ttl_valid_count_dict.values())
+    count_not_to_be_validated = sum(ttl_invalid_count_dict.values())
+
+    row_format ="{:<{col_size}} | {:<6} | {:<17}"
```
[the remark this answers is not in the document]
nmj101: added padding to valid directory lists to match beginning of data line of the warning list

scripts/m18_submission_checker/aida_m18_submission_checker.py
nmj101: just swapping them would mean you get a success when you expect partial success for the case of the missing NIST but existing INTER-TA

nmj101: believe i fixed this issue

scripts/m18_submission_checker/aida_m18_submission_checker.py
```diff
@@ -47,6 +55,9 @@ def get_valid_archive_filetype(archive_file_name):
     :rtype: string, or None
     """
 
+    # note - dont need to check for multiple occurances of file extentions, this will cause the get_task_type to fail
+    # with unkown task type.
```
craigwarsaw: mispelled "unknown"

PR #239, opened 2019-06-27
AIDA-890:  Incorporate AIDA-889 into v1.0.x

nextcen-dgemoets:
Incorporated AIDA-889 (Python LDCTime typo fix) into v1.0.x.

### DIFF
--- a/python/aida_interchange/LDCTimeComponent.py
+++ b/python/aida_interchange/LDCTimeComponent.py
@@ -7,7 +7,7 @@ class LDCTimeType(Enum):
     ON = 1
     BEFORE = 2
     AFTER = 3
-    UNKONWN = 4
+    UNKNOWN = 4
 
 class LDCTimeComponent:
     """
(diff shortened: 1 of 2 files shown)


python/aida_interchange/LDCTimeComponent.py
```diff
@@ -51,4 +51,3 @@ def make_aif_time_component(self, g):
         self.add_literal(g, time_component, AIDA_ANNOTATION.month, self.month, XSD.gMonth)
         self.add_literal(g, time_component, AIDA_ANNOTATION.day, self.day, XSD.gDay)
         return time_component
-
```
jakevb: I am a little confused.  Did you remove the newline at the end of this file?

nextcen-dgemoets: > I am a little confused. Did you remove the newline at the end of this file?

I believe the previous file didn't have a newline (red diff).  I believe my version removed the last line-- the line which didn't have a newline.  I could try adding the line back in (with a newline) so the only apparent diff is that the old one doesn't have a newline but the new one does?

jakevb: That doesn't really matter.  Removing the line works just as well.  Thanks for the clarification.

PR #243, opened 2019-06-27
Aida 874

sharknc:
- Updated code to differentiate between Task 1a and Task 1b submissions
- Implemented Task 1b job submission solution. Each hypothesis will be submitted as a unique job 
- FIxed bug where the job name and resulting tar name were invalid for Task 2 and Task 3 submissions because code was using the stem of the path instead of the name
- Updated the infrastructure batch CloudFormation script to implement a retry strategy of 3. Also updated default values to use 4 nodes per job.


jakevb:
A README should be added to the batch-single folder as one is specified to exist according to the README located in the batch-init folder.

PR #244, opened 2019-07-01
AIDA-888:  shallow validation

nextcen-dgemoets:
The `--depth` option performs a “shallow” validation in which each SHACL rule (shape) only considers a subset of its target nodes.

Notes:
- Specifying `--depth=-1` doesn't get rejected and performs a regular validation.  This is a picocli quirk as -1 is the "default" value.  The same thing happens with `--abort`.  Maybe I'll fix this...thoughts?
--> **Update:  fixed this.**
- I left in a line of debugging so you can see shallow validation "in action."  Obviously I'll remove it prior to merge.
- I'm not sure what a good default depth value is.  Thoughts?
- Test in combination with `--abort`, `-p`, `--p2`, `--disk`, and whatever other options you see fit.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,7 +59,7 @@ validate N files or all files in a specified directory.
 To run the validator from the command line, run `target/appassembler/bin/validateAIF`
 with a series of command-line arguments (in any order) honoring the following usage:  <br>
 Usage:  <br>
-`validateAIF [-hov] [--ldc] [--nist] [--nist-ta3] [--pm] [--program] [--abort[=num]] [-d=DIRNAME] [-t=num] [--ont=FILE...]... [-f=FILE...]...`  <br>
+`validateAIF [-hov] [--ldc] [--nist] [--nist-ta3] [--pm] [--program] [--abort[=num]] [--depth[=num]] [-d=DIRNAME] [-t=num] [--ont=FILE...]... [-f=FILE...]...`  <br>
 
 | Switch | Description |
 | ----------- | ----------- |
@@ -69,6 +69,7 @@ Usage:  <br>
 |`--nist` | validate against the NIST restrictions |
 |`--nist-ta3` | validate against the NIST hypothesis restrictions (implies `--nist`) |
 |`--abort[=num]` | Abort validation after `[num]` SHACL violations (num > 2), or three violations if `[num]` is omitted. |
+|`--depth[=num]` | Perform shallow validation in which each SHACL rule (shape) is only applied to `[num]` target nodes, or 50 nodes if `[num]` is omitted (requires -t). |
 |`--pm` | Enable progress monitor that shows ongoing validation progress |
 |`--disk` | Use disk-based model for validating very large files |
 |`-o` | Save validation report model to a file.  `KB.ttl` would result in `KB-report.txt`. Output defaults to stderr. |
--- a/src/main/java/com/ncc/aif/ValidateAIFCli.java
+++ b/src/main/java/com/ncc/aif/ValidateAIFCli.java
@@ -73,11 +73,10 @@ public class ValidateAIFCli implements Callable<Integer> {
     private static final Logger logger = (Logger) (org.slf4j.LoggerFactory.getLogger(Logger.ROOT_LOGGER_NAME));
     // ViolationThreshold
     private static final String ABORT_PARAMETER_STRING = "Abort parameter";
-    private static final int DEFAULT_MAX_VIOLATIONS = -1;
     private static final int MINIMUM_MAX_VIOLATIONS = 3;
     // Depth
     private static final String DEPTH_PARAMETER_STRING = "Depth parameter";
-    private static final int DEFAULT_DEPTH = -1;
+    private static final int DEFAULT_DEPTH = 50;
     private static final int MINIMUM_DEPTH = 1;
 
     // Profiling
@@ -112,7 +111,8 @@ public class ValidateAIFCli implements Callable<Integer> {
 
     @Option(names = "--abort", description = "Abort validation after [num] SHACL violations (num > 2), or 3 violations if [num] is omitted.",
             paramLabel = "num", arity = "0..1", converter = MaxErrorConverter.class)
-    private int maxValidationErrors = DEFAULT_MAX_VIOLATIONS;
+    private int maxValidationErrors = Integer.MIN_VALUE; // Don't fail-fast by default
+
     private static class MaxErrorConverter implements CommandLine.ITypeConverter<Integer> {
         @Override
         public Integer convert(String value) {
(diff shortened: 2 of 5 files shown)


README.md
```diff
@@ -69,6 +69,7 @@ Usage:  <br>
 |`--nist` | validate against the NIST restrictions |
 |`--nist-ta3` | validate against the NIST hypothesis restrictions (implies `--nist`) |
 |`--abort[=num]` | Abort validation after `[num]` SHACL violations (num > 2), or three violations if `[num]` is omitted. |
+|`--depth[=num]` | Perform shallow validation in which each SHACL rule (shape) is only applied to `[num]` target nodes, or 50 nodes if `[num]` is omitted (requires -t). |
```
nmj101: omitting num will error during runtime and will result in files being skipped.

src/main/java/com/ncc/aif/ValidateAIFCli.java
```diff
@@ -118,6 +124,20 @@ public Integer convert(String value) {
...
     }
 
+    @Option(names = "--depth", description = "Perform shallow validation in which each SHACL rule (shape) is only applied to [num] target nodes, or 50 nodes if [num] is omitted (requires -t).",
+            paramLabel = "num", arity = "0..1", converter = DepthConverter.class)
+    private int depth = DEFAULT_DEPTH;
+    private static class DepthConverter implements CommandLine.ITypeConverter<Integer> {
+        @Override
+        public Integer convert(String value) {
+            try {
+                return "".equals(value) ? MINIMUM_DEPTH : Integer.parseInt(value);
```
nmj101: I believe this is where you intend to set the value to the default if num was omitted. The descriptions say the is 50,  but MINIMUM_DEPTH is set to 1.

README.md
nextcen-dgemoets: I'm not getting an error when `[num]` is omitted, but it is using 1 instead of 50, as you say below.  Will fix in an upcoming commit.  It was poorly/erroneously copied from `--abort` where 3 is both the minimum and the default.

src/main/java/com/ncc/aif/ValidateAIFCli.java
```diff
@@ -59,6 +59,7 @@
     static final String ERR_MISSING_FILE_FLAG = "Must use one of these flags: -f | -d";
     static final String ERR_TOO_MANY_FILE_FLAGS = "Can only use one of these flags: -f | -d";
     static final String ERR_SMALLER_THAN_MIN = "%s must be at least %d";
+    static final String ERR_DEPTH_REQUIRES_T = "--depth requires -t";
```
nmj101: If run with -t=1 and --depth, it does error out with this message. Probably does not need to be addressed though.

src/main/java/com/ncc/aif/ValidateAIFCli.java
nmj101: I believe this is where you intend to set the value to the default if num was omitted. The descriptions say the default is 50, but MINIMUM_DEPTH is set to 1.

nextcen-dgemoets: This was poorly/erroneously copied from `--abort` where 3 is both the minimum and the default.  Setting `MINIMUM_DEPTH` to 50 is insufficient because it rejects anything less than 50 (e.g., `--depth=49`).  I need to separate the minimum, default, and detection of whether or not you specified anything.  Will do so in a future commit.

nmj101: resubmitting for request changes

src/main/java/com/ncc/aif/ValidateAIFCli.java
nextcen-dgemoets: True... `-t` is smart enough to run single-threaded validator if `t=1`.  I went ahead and tweaked the error message.

src/main/java/com/ncc/aif/ValidateAIFCli.java
nextcen-dgemoets: This should be fixed now.  Let me know if you see any other errors.

src/main/java/com/ncc/aif/ValidateAIFCli.java
```diff
@@ -72,8 +73,12 @@
     private static final Logger logger = (Logger) (org.slf4j.LoggerFactory.getLogger(Logger.ROOT_LOGGER_NAME));
     // ViolationThreshold
     private static final String ABORT_PARAMETER_STRING = "Abort parameter";
-    private static final int DEFAULT_MAX_VIOLATIONS = -1;
```
ecurley-nextcentury: Why not just have made DEFAULT_MAX_VIOLATIONS = Integer.MIN_VALUE?

src/main/java/com/ncc/aif/ValidateAIFCli.java
```diff
@@ -106,7 +111,8 @@
 
     @Option(names = "--abort", description = "Abort validation after [num] SHACL violations (num > 2), or 3 violations if [num] is omitted.",
             paramLabel = "num", arity = "0..1", converter = MaxErrorConverter.class)
-    private int maxValidationErrors = DEFAULT_MAX_VIOLATIONS;
+    private int maxValidationErrors = Integer.MIN_VALUE; // Don't fail-fast by default
```
ecurley-nextcentury: I thought when validationErrorBatch was -1, that turned fail-fast off. Reading this makes me think that fail-fast will now always perform the violation count check. It just won't ever throw the exception.Let me know what I'm missing

nextcen-dgemoets: In `ValidateAIF` itself, `-1` turns off fail-fast, and still does.  In ValidateAIFCli, `-1` was being used to detect whether or not the user was trying to enable fail-fast.  However, I noticed that both `--abort` and `--depth` didn't handle it properly when a parameter of `-1` was specified by the user.  It disabled the feature instead of giving an error saying the parameter was too small.

Instead, I'm setting the variables (`maxValidationErrors` and `depth`) to `Integer.MIN_VALUE` and using that to determine if the user tried to enable the feature.  If they don't specify the flags, then the feature is not enabled.

ecurley-nextcentury: Okey dokey. Marking resolved.

src/main/java/com/ncc/aif/ValidateAIFCli.java
nextcen-dgemoets: I slightly changed the context of these constants to make it work more consistently with `--depth`.  The `DEFAULT_*` and `MINIMUM_*` constants now refer to the default and minimum values (respectively) of the `[num]` parameter for these switches.  They no longer convey anything about whether the feature is enabled or not.

In fail-fast, the default and the minimum are the same (3), so I use the minimum constant.  It could be clearer or more readable/consistent if I set both `MINIMUM_MAX_VIOLATIONS` and `DEFAULT_MAX_VIOLATIONS` to 3.

That said, there _could_ be an additional constant set to `Integer.MIN_VALUE` to use for detecting whether or not the flag was specified...maybe called `UNSPECIFIED_VALUE`.  WDYT?

ecurley-nextcentury: I don't really have a strong opinion either way. Your reasoning is sound. Marking resolved.

nextcen-dgemoets:
Added CLI test... @ecurley-nextcentury please review.  Sorry I missed it in the first place!

PR #246, opened 2019-07-09
AIDA-904: Remove JVM RAM specification from POM

nextcen-dgemoets:
Removed extraJvmArguments from pom;
Documented memory considerations in README;
Added a section to README for shallow validation (since it was lacking).

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,7 +137,7 @@ To enable shallow validation programmatically, use `ValidateAIF.setDepth()` and 
 
 Validation of a large files can require significant system resources, particularly system RAM.  By default, the Java
 heap will use up to half of the available RAM on your system.  If you want to set a higher (or lower) maximum, then
-set the `JAVA_OPTS` environment variable to, for example, `-mx16G` to use up to 16GB of RAM.
+set the `JAVA_OPTS` environment variable to, for example, `-Xmx16G` to use up to 16GB of RAM.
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -121,6 +121,27 @@ Without the `--abort` option, the entire KB will be validated with full output o
...
+be tested.  The `--depth` option requires enabling the multi-threaded validator via the `-t` option.
+Unlike failing fast, shallow validation ends early whether or not it finds any SHACL violations.
+
+To enable shallow validation programmatically, use `ValidateAIF.setDepth()` and specify a depth.
+
+### Memory considerations
+
+Validation of a large files can require significant system resources, particularly system RAM.  By default, the Java
+heap will use up to half of the available RAM on your system.  If you want to set a higher (or lower) maximum, then
+set the `JAVA_OPTS` environment variable to, for example, `-mx16G` to use up to 16GB of RAM.
```
craigwarsaw: Shouldn't `-mx16G` be `-Xmx16G` ?

nextcen-dgemoets: According to [StackOverflow](https://stackoverflow.com/questions/5374455/what-does-java-option-xmx-stand-for), options which start with -X are non standard across platforms and may be removed in the future. There is a -mx which is standard, shorter and does the same thing.

That said, I'd never heard of it, and have always used -Xmx.  It's probably better-known overall.  Would you rather I switch to -Xmx?  I don't feel strongly either way.

nextcen-dgemoets: Fixed.

PR #265, opened 2019-10-14
AIDA-1074: Bump to 1.1.0-SNAPSHOT

craigwarsaw:


craigwarsaw:
> I wonder if we should call it the 1.0.5-SNAPSHOT, though.

I thought about that, but I think we already agreed that moving from JDK 8 to OpenJDK should bump the release number to 1.1.0, especially if that causes it to be incompatible with JDK 8.

PR #266, opened 2019-10-21
AIDA-1114 Hotfix

sharknc:
- Upgraded the com.fasterxml.jackson libraries to version 2.10.0 to address potential vulnerability.
- Changed AIF version to 1.0.5 in both Java and Python versions
- Updated all documentation to point to AIF 1.0.5

**After this gets merged into master we will need to tag it as release 1.0.5**

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -30,11 +30,11 @@ added.
 ### How do I migrate from the GAIA/Kotlin version of AIF to the Java version?
 
 First, if you haven't already, you'll need to change your build script or tool to import
-`aida-interchange-1.0.4.jar` (changed from `gaia-interchange-kotlin-1.0-SNAPSHOT.jar`).
+`aida-interchange-1.0.5.jar` (changed from `gaia-interchange-kotlin-1.0-SNAPSHOT.jar`).
 For gradle, for example, include the following in the dependencies in your `build.gradle` file:
 
 `dependencies {
-    compile 'com.ncc:aida-interchange:1.0.4'
+    compile 'com.ncc:aida-interchange:1.0.5'
 }`
 
 Next, update any source files that import the `edu.isi.gaia` package to use the `com.ncc.aif` package instead.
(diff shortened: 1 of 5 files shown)


FAQ.md
```diff
@@ -30,11 +30,11 @@ added.
 ### How do I migrate from the GAIA/Kotlin version of AIF to the Java version?
 
 First, if you haven't already, you'll need to change your build script or tool to import
-`aida-interchange-1.0.4.jar` (changed from `gaia-interchange-kotlin-1.0-SNAPSHOT.jar`).
+`aida-interchange-1.0.5.jar` (changed from `gaia-interchange-kotlin-1.0-SNAPSHOT.jar`).
```
nextcen-dgemoets: I believe that this content should be updated simply to point to the AIF 1.0.4 FAQ like it currently does in the develop branch.  It looks like this was supposed to be done as part of AIDA-1074 after the 1.0.4 tag was created.

And based on AIDA-1074, it may be that we're supposed to change the versions to `1.1.0-SNAPSHOT` (OR `1.1.0` for python) after v1.0.5 is made.  But we can discuss later.

sharknc: @nextcen-dgemoets not sure I am following. Should I revert it back to recommend importing the aida-interchange-1.0.4.jar?

nextcen-dgemoets: No, I'm saying I believe we should have the `How do I migrate from the GAIA/Kotlin version of AIF ...` question just to link to the AIF 1.0.4 FAQ like it does in the develop branch.  Let's verify with @craigwarsaw as well as what the versions should be after v1.0.5 is tagged.

craigwarsaw: Good questions and good conversation.  Before we decided to release an AIF 1.0.5 (when we thought that the next release was going to 1.1.x), I thought that this FAQ answer could freeze pointing to AIF 1.0.4.
After seeing how this is playing out, I suggest that we leave this as it is now (with the AIF 1.0.5 FAQ answer updated to reference 1.0.5).  In AIF 1.1.0-SNAPSHOT, let's just completely remove this FAQ entry.

PR #267, opened 2019-10-21
AIDA-1114

sharknc:
- Upgraded the com.fasterxml.jackson libraries to version 2.10.0 to address potential vulnerability.
- Updated FAQ documentation to point to AIF 1.0.5

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -26,7 +26,3 @@ relation 2 or relation 3") since we assume that such cases should expressed as t
 distinct hypotheses (e.g. one hypothesis for "relation 1 and relation 2" and another for
 "relation 1 and relation 3").  However, if expressing such disjunctions is desired, it can be
 added.
-
-### How do I migrate from the GAIA/Kotlin version of AIF to the Java version?
-
-Refer to the [AIF v1.0.4 FAQ](https://github.com/NextCenturyCorporation/AIDA-Interchange-Format/blob/tag/v1.0.4/FAQ.md)
--- a/FAQ.md
+++ b/FAQ.md
@@ -29,4 +29,4 @@ added.
 
 ### How do I migrate from the GAIA/Kotlin version of AIF to the Java version?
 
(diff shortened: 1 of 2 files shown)


FAQ.md
```diff
@@ -29,4 +29,4 @@ added.
 
 ### How do I migrate from the GAIA/Kotlin version of AIF to the Java version?
 
-Refer to the [AIF v1.0.4 FAQ](https://github.com/NextCenturyCorporation/AIDA-Interchange-Format/blob/tag/v1.0.4/FAQ.md)
\ No newline at end of file
+Refer to the [AIF v1.0.5 FAQ](https://github.com/NextCenturyCorporation/AIDA-Interchange-Format/blob/tag/v1.0.5/FAQ.md)
```
nextcen-dgemoets: I'm not 100% sure, but I think this is supposed to point to develop, not 1.0.5.
[AIDA-1102](https://nextcentury.atlassian.net/browse/AIDA-1102) states:
Link to AIF should specify branch, e.g., AIDA-Interchange-Format/tree/develop

That said, AIDA-1102 mentions `README.md`, `java/README.md`, and `python/README.md`, but does not mention `FAQ.md`.

Of course, even if it links to develop, what should the link text say?  Maybe just `AIF FAQ` without a version number?  We may need @craigwarsaw to weigh in.

nextcen-dgemoets: Maybe the thinking is that since this is just a documentation link, it should just point to the current latest release of AIF, which is about to be 1.0.5.  If that's the thinking, I can buy that and will approve the PR.

nextcen-dgemoets: Okay, so I think I finally remembered / figured this out.  In the FAQs moving forward (starting with v1.0.5), we are linking to the v1.0.4 FAQ for instructions on upgrading from Kotlin so we don't have to keep updating the version number in this file, since it's kind of pointless because it's doubtful anybody is still upgrading from kotlin to java.  Therefore, this file should not be incremented to v1.0.5 and should be left unchanged.

sharknc: I have reverted the change to this file.

PR #272, opened 2020-02-20
Aida 1242

nextcen-dgemoets:
Added `--debug` switch that enabled debug output (logging) to the multi-threaded validator.
When debugging is enabled, I suggest saving program output to a file and simplifying the output with `cut -d " " -f 1-2,6- filename.txt > cleanfile.txt`.

When testing, make sure to enable the multi-threaded validator with `-t`.

### DIFF
--- a/java/src/main/java/com/ncc/aif/ThreadedValidationEngine.java
+++ b/java/src/main/java/com/ncc/aif/ThreadedValidationEngine.java
@@ -35,7 +35,7 @@ import java.util.stream.Collectors;
  */
 public class ThreadedValidationEngine extends ValidationEngine {
     private static boolean initialized = false;
-    private static final Logger logger = (Logger) (org.slf4j.LoggerFactory.getLogger(Logger.ROOT_LOGGER_NAME));
+    private static final Logger logger = (Logger) (org.slf4j.LoggerFactory.getLogger(ThreadedValidationEngine.class));
     private static void initializeSHComponents() {
         if (!initialized) {
             FunctionRegistry.get().put(TOSH.hasShape.getURI(), ThreadSafeHasShapeFunction.class);
--- a/java/src/main/java/com/ncc/aif/ThreadedValidationEngine.java
+++ b/java/src/main/java/com/ncc/aif/ThreadedValidationEngine.java
@@ -209,7 +209,7 @@ public class ThreadedValidationEngine extends ValidationEngine {
 //        if (monitor != null) {
 //            monitor.beginTask("Validating " + rootShapes.size() + " shapes", rootShapes.size());
 //        }
-            logger.debug("Validating " + rootShapes.size() + " shapes.");
+            logger.debug("Validating {} shapes.", rootShapes.size());
             int i = 0;
             for (Shape shape : rootShapes) {
                 validationMetadata.add(executor.submit(getShapeTask(shape, i++, executor)));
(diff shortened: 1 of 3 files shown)


java/src/main/java/com/ncc/aif/ThreadedValidationEngine.java
```diff
@@ -207,7 +209,7 @@ private boolean exceedsMaximumNumberViolations(int violations) {
 //        if (monitor != null) {
 //            monitor.beginTask("Validating " + rootShapes.size() + " shapes", rootShapes.size());
 //        }
-
+            logger.debug("Validating " + rootShapes.size() + " shapes.");
```
ecurley-nextcentury: For the debug statements, please use parameterized messages: http://www.slf4j.org/faq.html#logging_performance

java/src/main/java/com/ncc/aif/ThreadedValidationEngine.java
```diff
@@ -310,14 +312,22 @@ public Resource getReport() {
...
+                            shape.getShapeResource().getLocalName() + ", d=" + (System.currentTimeMillis() - start));
+                }
 
                 List<RDFNode> filtered = focusNodeFilter != null ?
                         focusNodes.stream().filter(focusNodeFilter).collect(Collectors.toList()) :
                         focusNodes;
                 smd.filteredTargetCount = filtered.size();
+                if (smd.targetCount != smd.filteredTargetCount) {
+                    logger.debug("--> After filter, " + smd.filteredTargetCount + " node(s) remain.");
+                }
```
ecurley-nextcentury: Just create single log containing targetCount and filteredTargetCount

java/src/main/java/com/ncc/aif/ValidateAIFCli.java
```diff
@@ -338,6 +341,10 @@ public Integer call() {
             logger.info("-> Validation will use " + threads + " threads.");
             validator.setThreadCount(threads);
         }
+        if (debugOutput) {
+            logger.info("-> Multi-threaded validation debugging output enabled.");
+            ((Logger) (org.slf4j.LoggerFactory.getLogger(ThreadedValidationEngine.class))).setLevel(Level.DEBUG);;
+        }
```
ecurley-nextcentury: As the ThreadedValidationEngine is abstracted from what the CLI is aware of, I would like to see something in the ValidateAIF object that would enable debugging. It would be ideal if there was some way to combine this and the progress monitor into a single idea for ValidateAIF. To just get output, I'm fine with the current implementation, but if we're going to commit this, we should provide controls for this in the ValidateAIF object. Let me know if you want to discuss.

java/src/main/java/com/ncc/aif/ThreadedValidationEngine.java
nextcen-dgemoets: I didn't do this because I wanted to keep the logs as simple as possible and not include additional information unless warranted (that is, don't mention filter count unless something was filtered out).  Also, I guess I wanted to focus on the collection time, and wasn't sure how long the filtering might take and whether that should be included in the duration.

How about something like this:

`logger.debug("Collected {} target node(s) {}for {}, d={}", smd.targetCount,`
`(smd.targetCount == smd.filteredTargetCount) ? "" : "(" + smd.filteredTargetCount + " after filter) ",`
`shape.getShapeResource().getLocalName(), (System.currentTimeMillis() - start));`

Sorry, couldn't get formatting quite right.  Anyway, it only writes the filtered number if relevant, and should only incur the cost of the string concat if the `targetCount != filterCount`.

java/src/main/java/com/ncc/aif/ValidateAIFCli.java
nextcen-dgemoets: @craigwarsaw has said that he wanted this change committed as an undocumented command line option.  I agree that it would be nice to have the CLI class have no knowledge of the `ThreadedValidationEngine`.  Unfortunately it already does in how it refers to `ThreadedValidationEngine.SH_ABORTED` to see if the validation was aborted (although I see now that there's a `TODO` note to "come up with better property").

It just seems a little bit...misleading? to add a debugging parameter/flag/etc at the ValidateAIF if we have no intention of implementing anything for it other than in the `ThreadedValidationEngine`.  I also thought about having `--debug` require the `-t` flag, but that seemed like unnecessary CLI wrangling for a developer-only tool...but I could be convinced otherwise.

Happy to discuss any of this further.

java/src/main/java/com/ncc/aif/ThreadedValidationEngine.java
nextcen-dgemoets: Ah, yes-- thanks.  Fixed.

java/src/main/java/com/ncc/aif/ValidateAIFCli.java
nextcen-dgemoets: Perhaps I should also change the flag to `--debugMT`.

PR #273, opened 2020-03-25
AIDA-1309

nextcen-dgemoets:
- Improved `EntityShape` and `EventRelationShape` SPARQL query (courtesy of @ecurley-nextcentury).
- Added test for a cluster prototype missing a type assertion.
- Added a few `sh:messages` to certain SHACL rules.
- NOTE: this PR keeps the SHACL library version at v1.2.1.  [AIDA-1350](https://nextcentury.atlassian.net/browse/AIDA-1350) was created to upgrade to v1.3.2 once it's publicly released.  Alternatively, we can include v1.3.2-SNAPSHOT in the local repo.

### DIFF
--- a/java/src/main/resources/com/ncc/aif/aida_ontology.shacl
+++ b/java/src/main/resources/com/ncc/aif/aida_ontology.shacl
@@ -60,11 +60,9 @@ aida:EntityShape
             SELECT $this
             WHERE {
                 FILTER ( NOT EXISTS {
-                        ?typeAssertion a rdf:Statement .
-                        ?typeAssertion rdf:predicate rdf:type .
-                        ?typeAssertion rdf:subject $this .
-                    } && ( NOT EXISTS { ?membership aida:clusterMember $this . } ||
-                            EXISTS { ?cluster aida:prototype $this . } )
+                        $this ^rdf:subject / rdf:predicate rdf:type .
+                    } && ( NOT EXISTS { $this ^aida:clusterMember ?membership . } ||
+                           EXISTS { $this ^aida:prototype ?cluster . } )
                 )
             }
         """ ;
@@ -122,11 +120,9 @@ aida:EventRelationShape
             SELECT $this
             WHERE {
                 FILTER ( NOT EXISTS {
-                        ?typeAssertion a rdf:Statement .
-                        ?typeAssertion rdf:predicate rdf:type .
-                        ?typeAssertion rdf:subject $this .
-                    } && ( NOT EXISTS { ?membership aida:clusterMember $this . } ||
-                            EXISTS { ?cluster aida:prototype $this . } )
+                        $this ^rdf:subject / rdf:predicate rdf:type .
+                    } && ( NOT EXISTS { $this ^aida:clusterMember ?membership . } ||
+                           EXISTS { $this ^aida:prototype ?cluster . } )
                 )
             }
         """ ;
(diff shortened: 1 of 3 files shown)


java/src/main/resources/com/ncc/aif/aida_ontology.shacl
```diff
@@ -60,11 +60,9 @@ aida:EntityShape
...
             WHERE {
                 FILTER ( NOT EXISTS {
-                        ?typeAssertion a rdf:Statement .
-                        ?typeAssertion rdf:predicate rdf:type .
-                        ?typeAssertion rdf:subject $this .
-                    } && ( NOT EXISTS { ?membership aida:clusterMember $this . } ||
-                            EXISTS { ?cluster aida:prototype $this . } )
+                        $this ^rdf:subject / rdf:predicate rdf:type .
+                    } && ( NOT EXISTS { $this ^aida:clusterMember ?membership . } ||
+                           EXISTS { $this ^aida:prototype ?cluster . } )
```
ecurley-nextcentury: It doesn't have to happen, but should this sparql restriction be made into a common component that both the EntityShape and EventRelationShape share?

java/src/test/java/com/ncc/aif/ExamplesAndValidationTest.java
```diff
@@ -1355,6 +1356,25 @@ void relationOfUnknownType() {
             utils.testInvalid("Invalid: relation of unknown type");
         }
 
+        @Test
+        void clusterPrototypeMissingTypeAssertion() {
```
ecurley-nextcentury: I'm not sure how this is not covered by entityMissingType(). I agree that it's not explicitly covering this case, but logically, I think it does. Can you think of a situation in which entityMissingType() would pass and clusterPrototypeMissingTypeAssertion would fail?

nextcen-dgemoets: I agree that if a cluster prototype is missing a type assertion, then `entityMissingType()` will also fail, even though `entityMissingType()` isn't testing entity clusters.

I think I was being especially conservative/explicit (and arguably redundant) in this test because the impact of the query change is so immense.  I wanted to know beyond a shadow of a doubt that it was perfectly valid, by explicitly covering all (possible) cases in the truth table.

But that said, if you prefer that the test be removed, I will do so.  It will not catch anything that `entityMissingType()` won't catch.

java/src/main/resources/com/ncc/aif/aida_ontology.shacl
nextcen-dgemoets: It's your query-- if you feel better about using a common component, then that's fine with me.  I will want to re-test at least one KB for procedural reasons, but that doesn't take long.

ecurley-nextcentury: Do you know how to implement this as a shared component? If not, I'm good to leave it the way it is.

java/src/test/java/com/ncc/aif/ExamplesAndValidationTest.java
ecurley-nextcentury: Yeah, if that's the case, I would recommend removing the test. If you feel strongly that it conveys something to the researchers, we can leave it in

java/src/main/resources/com/ncc/aif/aida_ontology.shacl
nextcen-dgemoets: Let's leave it the way it is then.

java/src/test/java/com/ncc/aif/ExamplesAndValidationTest.java
nextcen-dgemoets: All things being equal, I would have left it in, but I don't feel strongly about it, nor does it convey anything important to the researchers.  The important thing is that we've validated that the new query work in all cases.  Test removed.

PR #278, opened 2020-03-31
AIDA-1354b: Bumped develop to v1.1.1-SNAPSHOT

nextcen-dgemoets:
Bumped develop to v1.1.1-SNAPSHOT.

### DIFF
--- a/python/setup.py
+++ b/python/setup.py
@@ -1,7 +1,7 @@
 from setuptools import setup, find_packages
 
 setup(name='aida-interchange',
-      version='1.1.0',
+      version='1.1.1',
       author='Ryan Gabbard',
       author_email='gabbard@isi.edu',
       description='AIDA Interchange Format tools',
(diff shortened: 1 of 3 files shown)


python/setup.py
```diff
@@ -1,7 +1,7 @@
 from setuptools import setup, find_packages
 
 setup(name='aida-interchange',
-      version='1.1.0',
+      version='1.1.1',
```
ecurley-nextcentury, line 4: Should we add '.dev1' to the end of this like we do in [VERDI](https://github.com/NextCenturyCorporation/VERDI/blob/develop/python/setup.py)?

nextcen-dgemoets: We didn't previously in AIF, but maybe this was something that was figured out after we had bumped to 1.1.0 in AIF python.  It looks like Verdi has its own version system (which I vaguely recall).

I seem to recall that v1.1.0-SNAPSHOT didn't work, so we didn't do it.  But again, maybe the .dev1 part was figured out later.

I'll add it to the AIF setup.py.

PR #279, opened 2020-04-08
AIDA-1372: Add example to generate LDCTime ranges

craigwarsaw:
The example test added in this branch generates the AIF shown on the "Support for Composite Cluster Member Information" to be made public to teams after review.
Note - I set the DUMP_ALWAYS and DUMP_TO_FILE constants to true, so you can easily find this generated example in java/target/test-dump-output/ExamplesAndValidationTest_ValidExamples_createEventsWithLDCTimeRanges.ttl
Before it gets merged in, these constants should be changed back to false.

### DIFF
--- a/java/src/main/java/com/ncc/aif/AIFUtils.java
+++ b/java/src/main/java/com/ncc/aif/AIFUtils.java
@@ -1507,7 +1507,10 @@ public class AIFUtils {
      * @param system The system object for the system which marks the time
      * @return
      */
-    public static Resource markLDCTimeRange(Model model, Resource toMark, LDCTimeComponent startEarliest, LDCTimeComponent startLatest, LDCTimeComponent endEarliest, LDCTimeComponent endLatest, Resource system) {
+    public static Resource markLDCTimeRange(Model model, Resource toMark,
+                                            LDCTimeComponent startEarliest, LDCTimeComponent startLatest,
+                                            LDCTimeComponent endEarliest, LDCTimeComponent endLatest,
+                                            Resource system) {
         final Resource ldcTime = makeAIFResource(model, null, AidaAnnotationOntology.LDC_TIME_CLASS, system);
         if (endLatest != null) {
             ldcTime.addProperty(AidaAnnotationOntology.LDC_TIME_END, endLatest.makeAIFTimeComponent(model));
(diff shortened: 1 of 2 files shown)


nextcen-dgemoets:
Was there any talk of validating that startEarliest  < startLatest or anything like that?

java/src/main/java/com/ncc/aif/AIFUtils.java
```diff
@@ -1495,6 +1495,37 @@ public static Resource markLDCTime(Model model, Resource toMark, LDCTimeComponen
...
+     * @param model  The underlying RDF model for the operation
+     * @param toMark The Event or Relation to add the LDC time data to
+     * @param startEarliest  {@link LDCTimeComponent} containing the earliest start time in the range
+     * @param startLatest    {@link LDCTimeComponent} containing the latest start time in the range
+     * @param endEarliest    {@link LDCTimeComponent} containing the earliest end time in the range
+     * @param endLatest      {@link LDCTimeComponent} containing the latest end time in the range
+     * @param system The system object for the system which marks the time
+     * @return
+     */
+    public static Resource markLDCTimeRange(Model model, Resource toMark, LDCTimeComponent startEarliest, LDCTimeComponent startLatest, LDCTimeComponent endEarliest, LDCTimeComponent endLatest, Resource system) {
```
nextcen-dgemoets, line 1510: I'd suggest a line break here.

craigwarsaw:
> Was there any talk of validating that `startEarliest` < `startLatest` or anything like that?

The API would be able to validate it, but the AIF validator may not.  We didn't promise that we could validate this.  If we can, great and @ecurley-nextcentury might know how.  There actually is nothing stopping you (in base AIF) from adding any number of LDCTimeComponents (although that may be restricted in NIST restricted AIF), so you wouldn't necessarily know what to validate.

java/src/main/java/com/ncc/aif/AIFUtils.java
craigwarsaw: Fixed

craigwarsaw:
> > Was there any talk of validating that `startEarliest` < `startLatest` or anything like that?
> 
> The API would be able to validate it, but the AIF validator may not. We didn't promise that we could validate this. If we can, great and @ecurley-nextcentury might know how. There actually is nothing stopping you (in base AIF) from adding any number of LDCTimeComponents (although that may be restricted in NIST restricted AIF), so you wouldn't necessarily know what to validate.

@nextcen-dgemoets I created AIDA-1375 for this.

nextcen-dgemoets:
Nope, it's good-to-go...merge away.

PR #280, opened 2020-05-28
AIDA-1459: Initial commit standalone AIF validator container image

craigwarsaw:
This AIF validator container is adapted from the existing batch-init / batch-single containers but is meant to be run locally (i.e., not in AWS Batch).  This is intended to be inserted into the evaluation pipeline after an algorithm runs and before the finalizer, however, I think there may be stand-alone uses for it.
Currently, all parameters are passed in as environment variables.  Arguably, passing them in as command-line variables to the ENTRYPOINT may be (more? or less?) useful.  
Feedback on any of this is welcome.

### DIFF
--- a/docker/aif-validator/Dockerfile
+++ b/docker/aif-validator/Dockerfile
@@ -39,24 +39,20 @@ ENV VALIDATION_HOME /opt/aif-validator
 RUN mkdir $VALIDATION_HOME
 
 # Define default arguments if nothing is supplied using --build-args
-ARG GIT_REPO
-ARG VALIDATOR_REPO
-ARG VALIDATOR_BRANCH
+ARG GIT_REPO=github.com
+ARG VALIDATOR_REPO=NextCenturyCorporation/AIDA-Interchange-Format.git
+ARG VALIDATOR_BRANCH=master
 
 # Pull Git Repository
-WORKDIR /tmp
-RUN git clone https://$GIT_REPO/$VALIDATOR_REPO -b $VALIDATOR_BRANCH
+WORKDIR $VALIDATION_HOME
+RUN git clone https://$GIT_REPO/$VALIDATOR_REPO -b $VALIDATOR_BRANCH .
 
 # Run Maven Build
-WORKDIR /tmp/AIDA-Interchange-Format/java
+WORKDIR $VALIDATION_HOME/java
 RUN mvn clean -Dmaven.test.skip=true package
-
-# Copy repo to aif validation directory
-RUN cp -R /tmp/AIDA-Interchange-Format/* $VALIDATION_HOME
 
 # clean up set working directory to home
 RUN yum clean all && \
-    rm -rf /tmp/AIDA-Interchange-Format && \
     rm -rf /var/cache/yum
 WORKDIR $HOME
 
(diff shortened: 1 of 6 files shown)


craigwarsaw:
> @craigwarsaw ,
> 
> Looks good. I had no issues following the readme. I was able to build the docker and docker run with a TA1 and TA2 VALIDATION_FLAG.

Thanks @phile-caci. @ecurley-nextcentury let me know if you would like to look at it before it gets merged to the develop branch.

ecurley-nextcentury:
> @ecurley-nextcentury let me know if you would like to look at it before it gets merged to the develop branch.

Yeah, I'll take a look at it after our morning meeting. Thanks!

docker/aif-validator/Dockerfile
```diff
@@ -0,0 +1,65 @@
...
+# Install AIF Interchange Format
+#------------------------------------------------------------- 
+# Create directory structure
+ENV VALIDATION_HOME /opt/aif-validator
+RUN mkdir $VALIDATION_HOME
+
+# Define default arguments if nothing is supplied using --build-args
+ARG GIT_REPO
+ARG VALIDATOR_REPO
+ARG VALIDATOR_BRANCH
```
ecurley-nextcentury, line 44: Did you want to move the defaults here, rather than define them in build.sh?

docker/aif-validator/Dockerfile
```diff
@@ -0,0 +1,65 @@
...
+# Pull Git Repository
+WORKDIR /tmp
+RUN git clone https://$GIT_REPO/$VALIDATOR_REPO -b $VALIDATOR_BRANCH
+
+# Run Maven Build
+WORKDIR /tmp/AIDA-Interchange-Format/java
+RUN mvn clean -Dmaven.test.skip=true package
+
+# Copy repo to aif validation directory
+RUN cp -R /tmp/AIDA-Interchange-Format/* $VALIDATION_HOME
```
ecurley-nextcentury, line 55: It looks like the intent here is to remove the AIDA-Interchange-Format folder. I _think_ that should be possible by specifying '.' as the destination for the clone.

craigwarsaw: @ecurley-nextcentury - I followed what @sharknc did in the batch-single docker.  But to clarify what you are suggesting... Are you saying that if I cloned it into the right directory to start with, there would be no reason to remove it a couple lines later?  Like:
```
# Pull Git Repository
WORKDIR $VALIDATION_HOME
RUN git clone https://$GIT_REPO/$VALIDATOR_REPO -b $VALIDATOR_BRANCH .

# Run Maven Build
WORKDIR $VALIDATION_HOME/java
RUN mvn clean -Dmaven.test.skip=true package

# clean up set working directory to home
RUN yum clean all && \
    rm -rf /var/cache/yum
WORKDIR $HOME
```

ecurley-nextcentury: Assuming I'm not missing some other reason to build it under tmp and then move it, yes I'm suggesting we put it in $VALIDATION_HOME when we clone it

docker/aif-validator/Dockerfile
craigwarsaw: Good idea - done.  Also added comments to build.sh.example to show how to override defaults.

docker/aif-validator/Dockerfile
craigwarsaw: I made the change you suggested.  I'm not sure why @sharknc did it that way, but it could be because there some churn in AIF when he was developing his containers.  I left batch-single as is to minimize testing for this change, as we don't plan to use batch validation for phase 2.

PR #281, opened 2020-06-03
Add optional uri_ref to link_to_external_kb

rogerbock:
This is allows a user to avoid blank nodes, which lead to non-determinism in the output


rogerbock:
My editor trimmed trailing whitespace, so if you want to see the substantive change, you should hide whitespace changes.

This change should have no effect on existing code, as the argument is optional and without it the behavior is unchanged.

ecurley-nextcentury:
@rogerbock, Thanks for this addition and sorry for the late reply. The changes look good, but we're using master exclusively for the latest release. Can you create a pull request against the `develop` branch?

rogerbock:
Per @ecurley-nextcentury's request, closing this in favor of a pull request against the develop branch (#283)

PR #283, opened 2020-06-23
Add optional uri_ref argument to link_to_external_kb

rogerbock:


rogerbock:
If you hide the whitespace changes (my IDE stripped off trailing whitespace on lines), you will actually see the small, substantive change.

PR #305, opened 2021-11-16
AIDA-1976 

gitnordt:
This python script allows for any number of example "Claim Frame" ttl files to be created. Run in a virtual python environment. The output directory that will contain the ttls will be created when you run the script: python claims_creation.py


phile-caci (approved):
Moved the claims_creation.py to verdi-system\data\claim-frames
Updated claimIds per Hoa's email

PR #309, opened 2021-12-06
Aida 1443

phile-caci:
- Updates to default --disk, --disk removed since it is now the default.
- --mem for mem model
- changed additional --ldc ontology defaults to --dwd


gitnordt (approved):
Looks good!

PR #314, opened 2022-01-31
added - Associated KEs must have cluster >=1

omitz:
Added shacl support for  2.4 #16. The graph of Associated KEs must have at least one event cluster or relation cluster with at least one edge.


phile-caci (approved):
Looks good, tested against all of LDC's Ta3 Annotations and found no issues. Approved.

PR #316, opened 2022-02-02
Aida 2054 - Valid predicates for Event/Relation

phile-caci:
Hello @omitz ,

Nothing too exciting here.  I wanted to pull in the list of valid Event/Relation predicates but wasn't sure of the best way to do it.  I created both a small ontology to hold the valid predicates as well as a file with some 'oneof' statements that we can use in shacl if we had to.

It's not anything that will effect the current validations.  Just some pre-work for getting valid predicates.

Let me know if you have any questions.

Thanks,
Phi


java/src/main/java/com/ncc/aif/ValidateAIF.java
```diff
@@ -188,7 +189,7 @@ public static ValidateAIF createForDWD(Restriction restriction) {
                 restrictions.add(NIST_CLAIMFRAME_SHACL_RESNAME);
             case NIST:
                 restrictions.add(NIST_SHACL_RESNAME);
-                restrictions.add(NIST_CLAIMFRAME_SHACL_RESNAME);
+                //restrictions.add(NIST_CLAIMFRAME_SHACL_RESNAME);
```
omitz, line 192: Maybe just remove this commented line if we don't need it.

omitz (approved):

PR #323, opened 2022-03-01
Changed AIF to use memory model as default instead of disk model

omitz:
Changed AIF to use memory model as default instead of disk model


phile-caci (approved):
Approved... heh.. glad it was simple.  Thanks!

PR #326, opened 2022-03-09
Aida 2078

omitz:
This change validates the all files within all subdirectories of a given directory.


phile-caci (approved):
Tested, looks good.  Also removed testing of claims for now since Edge fixed breaks the test.  TestUtils is written to take triples as resource, resource, resource.  But AssociatedKE's sameascluster's prototype's arugment statement's predicate can be a string (a string from working ontology group). Another ticket will be created to fix testUtils

PR #332, opened 2022-04-06
Update subject prototype check for AssociatedKE's Event/Relation prot…

omitz:
This is in reference to:  "16. The graph of Associated KEs must have at least one event cluster or relation cluster with at least one edge" - Eval Plan

Hoa - It’s weird but not invalid for :associatedKEs to include an event cluster whose prototype is the object of an argument statement, without including the event or relation whose prototype is the subject of that argument statement:

cmuta1:corpora\/eval\/role\-stmt\-LhR25oJrfRzYrqttF2SEhn rdf:subject cmuta1:corpora\/eval\/event\-instance\-L0C049DU0\-r202201290059\-118 .

Remove Subject, as long as the prototype is in either of the subject or object of the argument statement then it should pass validation


phile-caci (approved):
Looks good, thanks!

